PR #4, opened 2021-03-11
Benchmark for committing a tree with 1000 leaves

s1na:
(Comment updated with more comprehensive results, see below)

The result I'm getting:

```bash
❯ go test -v -run=BenchmarkCommit1kLeaves -tags bignum_hbls -bench=.
...
BenchmarkCommit1kLeaves/insert-8                       1        1451244411 ns/op        14297256 B/op      11412 allocs/op
BenchmarkCommit1kLeaves/insertOrdered-8                1        1486191367 ns/op        14445392 B/op      14381 allocs/op
```

And:

```bash
❯ go test -v -run=BenchmarkCommit1kLeaves -tags bignum_kilic -bench=.
BenchmarkCommit1kLeaves/insert-8                       2         705902770 ns/op        81831024 B/op     561279 allocs/op
BenchmarkCommit1kLeaves/insertOrdered-8                2         742656540 ns/op        84141580 B/op     576989 allocs/op
PASS
```

---

 UPDATE: We noticed that using LinCombG1 as per #3 vs the previous [implementation](https://github.com/gballet/go-verkle/blob/32361a96839c506fd794bf2565e94de53960923e/tree.go#L318-L328) (let's call it baseline) has a significant performance difference, with baseline often showing a speed-up of 2-3x. The main slowdown in LinCombG1 seems to stem from empty nodes. As suggested by @gballet I then benchmarked committing to one full node with 1024 children (i.e. no empty nodes) and there LinCombG1 does indeed perform better. Therefore the results include committing to a tree with 1000, 10000, and 100000 leaves, as well as a full node with 1024 leaves. Since runtime of the `Insert` and `InsertOrdered` approaches are similar I only include `Insert` results.:
 
|  | 1k kilic | 1k hbls | 10k kilic | 10k hbls | 100k kilic | 100k hbls | FullNode kilic | FullNode hbls |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| LinCombG1 (s) | 0.7  | 1.45 | 4 | 7 | 20 | 36 | 0.02 | 0.04 |
| Baseline (s)  | 0.14 | 0.12 | 1.3 | 1 | 13 | 9 | 0.11 | 0.08 |

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -32,7 +32,6 @@ import (
 	"github.com/protolambda/go-kzg/bls"
 )
 
-
 func calcR(cs []*bls.G1Point, indices []*bls.Fr, ys []*bls.Fr) bls.Fr {
 	digest := sha256.New()
 	for _, c := range cs {
--- a/tree_test.go
+++ b/tree_test.go
@@ -28,6 +28,8 @@ package verkle
 import (
 	"bytes"
 	"testing"
+    "math/rand"
+    "time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/protolambda/go-kzg"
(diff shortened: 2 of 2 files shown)


s1na:
If I revert #3 there's a significant speed boost:

```bash
❯ go test -v -run=BenchmarkCommit1kLeaves -tags bignum_hbls -bench=.
BenchmarkCommit1kLeaves/insert-8                       9         120960711 ns/op        13569550 B/op      13704 allocs/op
BenchmarkCommit1kLeaves/insertOrdered-8                8         140465323 ns/op        13706397 B/op      16460 allocs/op
```

And:

```bash
❯ go test -v -run=BenchmarkCommit1kLeaves -tags bignum_kilic -bench=.
BenchmarkCommit1kLeaves/insert-8                       7         147013777 ns/op        25833409 B/op      79954 allocs/op
BenchmarkCommit1kLeaves/insertOrdered-8                6         175837915 ns/op        28316461 B/op      96798 allocs/op
```

s1na:
Added another benchmark which computes the commitment for one full node with 1024 children (i.e. no empty nodes):

1. Using LinCombG1 (as per #3):
  - hbls: 0.04s
  - kilic: 0.02s
2. Not using LinCombG1:
  - hbls: 0.08s
  - kilic: 0.11s

PR #5, opened 2021-03-15
Add bench for inserting 10k values in tree with 1M values

s1na:
The result of 10 runs is as follows. Note we're not using LinCombG1 (i.e. manually reverted #3) and bignum lib is hbls.

```
Took 88.370121ms to insert and commit 10000 leaves
Took 1.000353643s to insert and commit 10000 leaves
Took 89.519466ms to insert and commit 10000 leaves
Took 167.110938ms to insert and commit 10000 leaves
Took 94.143657ms to insert and commit 10000 leaves
Took 74.528575ms to insert and commit 10000 leaves
Took 31.913447ms to insert and commit 10000 leaves
Took 81.51628ms to insert and commit 10000 leaves
```

### DIFF
--- a/benchs/main.go
+++ b/benchs/main.go
@@ -1,0 +1,92 @@
+package main
+
+import (
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/gballet/go-verkle"
+	"github.com/protolambda/go-kzg"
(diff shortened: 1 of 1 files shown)


gballet (approved):
LGTM

PR #6, opened 2021-03-15
Add bench for modifying 10k in tree of size 1M

s1na:
On my machine it takes on the order of 7ms to edit 10000 randomly selected keys in a tree with 1M leaves. Note results were collected after reverting #3. Bignum lib is hbls.

```
BenchmarkModifyLeaves-8              160           6957828 ns/op               0 B/op          0 allocs/op
```


gballet (approved):
LGTM

PR #7, opened 2021-03-17
Switch to LinComb when committing after a threshold

s1na:
I got to number 110 purely based on a few runs on my own machine


gballet (approved):
I was actually thinking of using reference counting in the node itself, but for now it doesn't really matter.

PR #12, opened 2021-04-01
Add serialization to nodes

s1na:
Encode nodes via RLP. Internal nodes are handled specially by encoding a bitlist of which children are present and then a concatenation of non-empty children hashes.

Note that now the node hashes and their serialization have diverged..we should probably take care of that.

Also added the `Flush` method to the interface so can be called from downstream

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -75,8 +75,6 @@ type VerkleNode interface {
 	// is from the bottom of the tree, up to the root.
 	GetCommitmentsAlongPath([]byte) ([]*bls.G1Point, []*bls.Fr, []*bls.Fr, [][]bls.Fr)
 
-	Flush(chan FlushableNode)
-
 	// Serialize encodes the node to RLP.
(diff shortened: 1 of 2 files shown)


gballet (approved):
LGTM. We were wondering if it really made sense to export `Flush` for non-internal nodes. ~~The fact is, that the channel can be used for other things than saving nodes to the DB, so supporting them makes sense after all.~~ <- scratch that, even though this feature is a nice to have, it requires an extra call to `Hash` and so far there is no practical use for this, so we chose to make `internalNode` public instead.

gballet:
> Note that now the node hashes and their serialization have diverged..we should probably take care of that.

Our serialization is an implementation detail, and so it shouldn't be reflected. It's already such a nightmare to handle in the MPT, because in order to get the correct root, you need to have a functioning RLP library. Let's keep things simple at interface level and optimize under the hood.

PR #15, opened 2021-04-02
Introduce TreeConfig to handle trees with different sizes

gballet:
This is so that 1024- and 256-width trees can easily be compared.

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -27,13 +27,14 @@ package verkle
 
 import (
 	"crypto/sha256"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/protolambda/go-kzg"
 	"github.com/protolambda/go-kzg/bls"
 )
 
-func calcR(cs []*bls.G1Point, indices []*bls.Fr, ys []*bls.Fr) bls.Fr {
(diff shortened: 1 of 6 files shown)


proof.go
```diff
@@ -80,90 +81,56 @@ func calcQ(e, d *bls.G1Point, y, w *bls.Fr) bls.Fr {
...
-		for i := 0; i < nodeWidth; i++ {
+		for i := 0; i < tc.nodeWidth; i++ {
 			bls.MulModFr(&tmp, &powR, &quotients[i])
 			bls.AddModFr(&g[i], &g[i], &tmp)
 		}
 
 		// rⁱ⁺¹ = r ⨯ rⁱ
 		bls.MulModFr(&powR, &powR, &r)
 	}
-	d = bls.LinCombG1(lg1, g[:])
```
s1na, line 154: I just noticed `LinCombG1` is being used here. Wondering if the same trick for switching algorithms could help here too

s1na (approved):
Looks good!

proof.go
gballet: most likely :+1:

gballet: Actually, let me take this back: `g` is calculated on the evaluation basis, so it's unlikely that any coefficient will be zero.

PR #19, opened 2021-04-13
sanity check in insert

gballet:


s1na (approved):
Looks good. `leaf.Insert` is being called by `leaf.InsertOrdered`, but I couldn't find where `leaf.InsertOrdered` itself is being called to this code might be right now unused.

gballet:
yeah, it shouldn't be called. Might as well panic in any case. This is just here in case we switch to an iterative approach, and the code will be incorrect in that case anyway. I'll mull it over.

gballet:
Closing, this is mixing `Insert` and `Update`.

PR #22, opened 2021-04-15
Make internal hash sha256 of compressed commitment

s1na:


tree.go
```diff
@@ -318,12 +319,9 @@ func (n *InternalNode) Get(k []byte) ([]byte, error) {
...
 
 func (n *InternalNode) Hash() common.Hash {
-	digest := sha256.New()
-	for _, child := range n.children {
-		digest.Write(child.Hash().Bytes())
-	}
-
-	return common.BytesToHash(digest.Sum(nil))
+	comm := n.ComputeCommitment()
+	h := sha256.Sum256(bls.ToCompressedG1(comm))
```
gballet, line 323: it would make sense to only compute the commitment if it's not already computed, otherwise it's computed twice in `InsertOrdered`.

```suggestion
	if n.commitment == nil {
		n.ComputeCommitment()
	}
	h := sha256.Sum256(bls.ToCompressedG1(n.commitment))
```

s1na: ComputeCommitment already has this condition and will only compute if its not cached

gballet (approved):
LGTM

tree.go
gballet: It's actually a big problem that `ComputeCommitment` only computes once, and then serves the cached version - because in the case of an update, it will still serve the old version. I'll create an issue to remember to fix that.

PR #25, opened 2021-04-16
Fix nodes not being flushed

s1na:


gballet (approved):
LGTM, feather in your hat for catching this one

PR #28, opened 2021-04-20
Use a singleton to obtain computing assets

gballet:
#15 introduced the possibility to switch between node sizes of 256 and 1024. Doing so, an `InitTreeConfig` function was created. The issue was that it was called every time a new tree would be created, and recalculate the lagrange basis, which is very time consuming.

This PR introduces the singleton method `GetTreeConfig` that will only calculate it once. This brings the root commitment calculation times back to its pre-#15 levels.

Crediting Kev in this PR, as it very partially implement an API change that he suggested.


s1na (approved):
LGTM

PR #29, opened 2021-04-21
Add node decoding support

s1na:
Distinguishes between RLP-serialized node types and decodes them into their respective objects. I'm planning to move encoding logic from this repo to go-ethereum soon.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -9,7 +9,7 @@ import (
 
 const ErrInvalidNodeEncoding = "invalid node encoding"
 
-func ParseNode(serialized []byte, tc *TreeConfig) (VerkleNode, error) {
+func ParseNode(serialized []byte, depth int, tc *TreeConfig) (VerkleNode, error) {
 	elems, _, err := rlp.SplitList(serialized)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


encoding.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return nil, err
+	}
+	c, err := rlp.CountValues(elems)
+	if err != nil {
+		return nil, err
+	}
+
+	if c == 1 {
+		// hashedNode
```
gballet, line 23: I don't think it makes sense to save a `hashedNode` as the data is already present in the db and we can't really do anything with them. I understand this is connected to `insertOrdered` hashing previous nodes and thus causing most children in a node to be `hashedNode`s. This comforts me in my opinion, expressed in #30, that we should change `insertOrdered` not to hash leaves, and not save them to db when they are flushed. Parent nodes + its leaves should be committed to db together.

encoding.go
```diff
@@ -0,0 +1,89 @@
...
+	elems, _, err := rlp.SplitList(serialized)
+	if err != nil {
+		return nil, err
+	}
+	c, err := rlp.CountValues(elems)
+	if err != nil {
+		return nil, err
+	}
+
+	if c == 1 {
```
gballet, line 22: I advise a `swtich`/`case` and also to define some constants instead of just having numbers.

encoding.go
```diff
@@ -0,0 +1,89 @@
...
+		// either leaf or internal
+		kind, first, rest, err := rlp.Split(elems)
+		if err != nil {
+			return nil, err
+		}
+		if kind != rlp.String {
+			return nil, errors.New(ErrInvalidNodeEncoding)
+		}
+
+		if len(first) == 32 {
```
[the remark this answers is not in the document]
gballet: also, why aren't you using the length of the first element when you could use the type to differentiate between a leaf and an internal node ? Not that we really care, I think leaves should be embedded in their parents so the point is moot.

gballet (commented):
Nice. It would be nice to add another test that tries to insert random keys, just so that we have more confidence in the code. A serialization and deserialization benchmark would also help.

encoding.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+const ErrInvalidNodeEncoding = "invalid node encoding"
+
+func ParseNode(serialized []byte, tc *TreeConfig) (VerkleNode, error) {
```
gballet, line 12: I suggest changing the signature of this function to `(serialized []byte, width int)` because then you can pass the result of calling `GetTreeConfig(width)` as a parameter to `createInternalNode`. This way, `TreeConfig` is that much closer to not being exported outside the `verkle` package.

encoding.go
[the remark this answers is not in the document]
gballet: ok, `c == number of elements in the top-level list`

gballet (approved):
LGTM. We decided not to change the signature as, in the future, we might want to pass extra parameters besides the `width`.

PR #30, opened 2021-04-21
serialize leaves into their parent

gballet:


s1na (approved):
Looks good

gballet:
This provides great speedups for the conversion, but comes with a penalty in execution mode, as much more data need to be loaded and saved to disk. Closing :cry:

Issue #16, opened 2021-04-02
Implement the unified trie


gballet:
Modify Verkle trees to support the scheme described [here](https://ethereum-magicians.org/t/proposed-verkle-tree-scheme-for-ethereum-state/5805)

gballet:
geth-side poc at https://github.com/ethereum/go-ethereum/pull/22999

gballet:
done

PR #31, opened 2021-04-21
Implement the merge of accounts + storage tries

gballet:
This is the go-verkle part of #16. It adds a new `accountLeaf` node that holds account data. `leafNode` is now used to represent storage slots and has to be 32-byte long.

Account nodes can't be written directly, they have to be created through the `CreateAccount` methods.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -56,6 +56,9 @@ type VerkleNode interface {
 	// is clear that no more values will be inserted in there.
 	InsertOrdered([]byte, []byte, chan FlushableNode) error
 
+	// Adds an account to the tree
+	CreateAccount([]byte, uint64, uint64, uint64, *big.Int, common.Hash) error
+
 	// Get value at key `k`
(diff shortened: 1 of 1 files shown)


tree.go
```diff
@@ -428,7 +481,139 @@ func (n *InternalNode) Serialize() ([]byte, error) {
...
+	// situation doesn't occur. This check will catch an
+	// invalid situation while the library stabilizes.
+	if !bytes.Equal(k[:31], n.key[:31]) {
+		return errors.New("inserting invalid key into key")
+	}
+
+	n.key = k
+	n.Version = version
+	n.Balance = balance
+	n.Nonce = nonce
```
gballet, line 516: note to self: does it make sense to set the nonce from the start? This function is trying to realize both "create" when an account is created, and "insert" when an account is loaded into the tree. For a prototype, it's fine. A note should be taken, though.

gballet:
superseded by #59, closing

PR #33, opened 2021-04-21
Fix commitment going stale after update

### LINKED ISSUE #23: Don't return cached commitment in ComputeCommitment.
As discovered [here](https://github.com/gballet/go-verkle/pull/22#discussion_r614055734), `ComputeCommitment` will cache the output, and calling it a second time will return the same value.

This is a problem, because when the tree is updated, it will still serve the old value instead of computing the new one. Furthermore, serving the cached version is the purpose of `GetCommitment` so there is no reason for this to happen in this function.

I'm opening an issue as a reminder to first ensure that this invalid behavior is not assumed anywhere in the code. One instance is that `Hash` will assume that.

s1na:
Fixes #23

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -175,6 +175,11 @@ func offset2KeyTenBits(key []byte, offset int) uint {
 }
 
 func (n *InternalNode) Insert(key []byte, value []byte) error {
+	// Clear cached commitment on modification
+	if n.commitment != nil {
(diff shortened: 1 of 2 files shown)


gballet (approved):
LGTM. There's a rider in the benchmark code, that I agree with as well.

PR #34, opened 2021-04-26
Naive multiproof implementation

gballet:
Change function signatures to accept multiple keys. The proof elements for each key are calculated independently, and then these results are naively concatenated.

In the implementation, only the  root will have multiple keys as an argument, and then each subtree will receive an array containing only one value, thus duplicating values for nodes having more than one child in the proof.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -68,11 +68,11 @@ type VerkleNode interface {
 	// commitment of a node.
 	GetCommitment() *bls.G1Point
 
-	// GetCommitmentAlongPath follows the path that one key
-	// traces through the tree, and collects the various
+	// GetCommitmentAlongPaths follows the path traced
+	// by keys through the tree, and collects the various
 	// elements needed to build a proof. The order of elements
 	// is from the bottom of the tree, up to the root.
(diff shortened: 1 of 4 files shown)


tree.go
```diff
@@ -443,9 +443,9 @@ func (n *InternalNode) GetCommitment() *bls.G1Point {
 	return n.commitment
 }
 
-func (n *InternalNode) GetCommitmentsAlongPath(key []byte) ([]*bls.G1Point, []*bls.Fr, []*bls.Fr, [][]bls.Fr) {
-	childIdx := Offset2Key(key, n.depth, n.treeConfig.width)
-	comms, zis, yis, fis := n.children[childIdx].GetCommitmentsAlongPath(key)
+func (n *InternalNode) getCommitmentsAlongPath(keys [][]byte) ([]*bls.G1Point, []*bls.Fr, []*bls.Fr, [][]bls.Fr) {
+	childIdx := Offset2Key(keys[0], n.depth, n.treeConfig.width)
+	comms, zis, yis, fis := n.children[childIdx].GetCommitmentsAlongPaths(keys)
```
s1na, line 448: I imagine this line is the reason this function is taking an array of keys? IMO it'd be cleaner to take one key and do `[][]byte{key}` here and avoid `keys[0]` above

gballet:
Done in 87d2e971, closing

PR #35, opened 2021-04-26
Add a method to copy trees

gballet:
`geth` is making copies of the tree every time a new transaction is executed, in the event that it is reverted. Introduce the `Copy()` helper function, that does this.

### DIFF
--- a/tree_test.go
+++ b/tree_test.go
@@ -367,8 +367,9 @@ func TestCopy(t *testing.T) {
 	tree.ComputeCommitment()
 
 	copied := tree.Copy()
+	copied.(*InternalNode).clearCache()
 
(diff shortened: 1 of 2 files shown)


tree_test.go
```diff
@@ -352,6 +352,32 @@ func TestFlush1kLeaves(t *testing.T) {
...
+	key3 := common.Hex2Bytes("0405000000000000000000000000000000000000000000000000000000000000")
+	tree := New(8)
+	tree.Insert(key1, value)
+	tree.Insert(key2, value)
+	tree.Insert(key3, value)
+	tree.ComputeCommitment()
+
+	copied := tree.Copy()
+
+	if !bytes.Equal(bls.ToCompressedG1(copied.GetCommitment()), bls.ToCompressedG1(tree.GetCommitment())) {
```
s1na, line 368: You can also compare `copied.Hash()` against `tree.Hash()`

PR #36, opened 2021-04-28
Add a node resolution callback to Get

gballet:
In order to perform #16, some form of node resolution method is needed when encountering a `HashedNode` in `Get`. This will recursively load nodes from the database. Thread-safety isn't guaranteed here, nonetheless it will work well for the prototype in geth.

~~:warning:  This embeds `encoding.go` from #29, which hasn't been merged yet and needs to be updated.~~ rebased on top of the latest `encoding.go`

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -1,3 +1,28 @@
+// This is free and unencumbered software released into the public domain.
+//
+// Anyone is free to copy, modify, publish, use, compile, sell, or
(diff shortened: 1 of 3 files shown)


s1na (approved):
So my only comment is that `Get` is now mutating the tree by replacing hashed nodes with their full representation, but I don't see how that should be a problem. Otherwise seems like a clean solution

PR #38, opened 2021-05-04
match geth's behavior: return nil on non-existing key

gballet:
In Geth, the `SecureTrie` will return `nil` when one tries to `Get` a value that is absent from the trie. This PR makes Verkle trees adopt the same behavior.


s1na (approved):
LGTM

PR #43, opened 2021-05-06
Add delete method

s1na:
An alternative is to do deletion when value passed to Insert is empty/nil.

Also should we return an error when trying to delete a non-existent key?

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -92,8 +92,9 @@ const (
 )
 
 var (
-	errInsertIntoHash  = errors.New("trying to insert into hashed node")
-	errValueNotPresent = errors.New("value not present in tree")
+	errInsertIntoHash    = errors.New("trying to insert into hashed node")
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -331,6 +334,32 @@ func (n *InternalNode) InsertOrdered(key []byte, value []byte, flush chan Flusha
...
+func (n *InternalNode) Delete(key []byte) error {
+	// Clear cached commitment on modification
+	if n.commitment != nil {
+		n.commitment = nil
+	}
+
+	nChild := Offset2Key(key, n.depth, n.treeConfig.width)
+	switch child := n.children[nChild].(type) {
+	case Empty:
+		// TODO: should return error?
```
gballet, line 346: yes, please return an error. Not that I believe this should be the final behavior, however it's mighty useful when debugging.

PR #44, opened 2021-05-06
Encoding fixes

s1na:
* Pass width as parameter to `ParseNode` instead of tree config
* Add type to the serialization of leaf and internal nodes


gballet (approved):
LGTM

PR #46, opened 2021-05-07
Fix deletion pruning

s1na:
Handles the cases where after deletion an internal node becomes empty and can be itself deleted, and the case where an internal node will have one remaining node which will in turn replace its parent.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -360,6 +360,9 @@ func (n *InternalNode) Delete(key []byte) error {
 		return nil
 	default:
 		child.Delete(key)
+		if len(child.(*InternalNode).nonEmptyIndices()) == 0 {
+			n.children[nChild] = Empty{}
+		}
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -360,6 +360,15 @@ func (n *InternalNode) Delete(key []byte) error {
 		return nil
 	default:
 		child.Delete(key)
+		// Prune child if necessary
+		nonEmpty := child.(*InternalNode).nonEmptyIndices()
```
gballet, line 364: I'm not a big fan of allocating memory just to get the size, here. Wouldn't it be possible to have one loop counting non-`nil` children, and bails if the count is greater than 2?

gballet (changes requested):

gballet (approved):

PR #47, opened 2021-05-07
copy: only copy commitments if they have been calculated in hash/internal

gballet:
Check that the commitments are not `nil` and if so, copy the commitment to the destination tree.

Another option would be to error, but so far in my geth experiments it seems that it's better not to recalculate the commitments each time that the trie is updated.


s1na (approved):
LGTM

PR #49, opened 2021-05-17
Remove unnecessary copy in hashToFr

gballet:
This is a small rewrite of `hashToFr` that doesn't allocate as much memory, even if most of the inefficiencies remain.


s1na (approved):
Logic is equivalent 👍

Issue #51, opened 2021-05-19
kililc give different commitments depending on the value of multiExpThreshold


gballet:
`multiExpThreshold == 45`:

```
INFO [05-18|19:42:21.287] Iterating state snapshot                 accounts=114215117 slots=400207177 elapsed=12h21m26.912s eta=2.205s
INFO [05-18|19:42:21.537] Iterated snapshot                        accounts=114215117 slots=400223042 elapsed=12h21m27.162s
INFO [05-18|19:42:21.537] Computed verkle tree                     root hash="542518…3afdb6"
INFO [05-18|19:42:21.539] Number of nodes written to DB
           nodes=174639610
```

`multiExpThreshold == 20`:
```
INFO [05-19|04:00:27.202] Iterating state snapshot                 accounts=114215117 slots=400208359 elapsed=15h45m18.981s eta=3.19s
INFO [05-19|04:00:27.785] Iterated snapshot                        accounts=114215117 slots=400223042 elapsed=15h45m19.564s
INFO [05-19|04:00:27.785] Computed verkle tree                     root hash="2c9607…dcf64f"
INFO [05-19|04:00:27.787] Number of nodes written to DB
           nodes=689077769
```

Looks the same number of nodes didn't get written to othe db, which certainly explains the time difference. Is it also the reason for obtaining different roots?

gballet:
actually, multiple runs with the same constant give different results, so there is an issue with our conversion code

gballet:
Just verified that herumi gives the proper result, it's kilic specific.

gballet:
Possible duplicate of #50, after all if kilic isn't consistent with itself, it can't be constantly consistent with herumi either.

gballet:
problem identified: kilic is not thread-safe.

gballet:
This issue has two parts:
 1. `GetTreeConfig` wasn't reentrant and the same initialization code could be run concurrently. This has been fixed in #53.
 2. Kilic isn't reentrant either, which has been reported upstream. Not much we can do about this for the moment, except keep #50 open as a tracking issue.

Closing this issue as it is now redundant with #50.

PR #53, opened 2021-05-19
guard configuration with a mutex

gballet:
Addresses #51 : Add a mutex in `GetTreeConfig` in order to avoid the thread-safety problems of kilic


s1na (approved):
Gonna approve, hoping it won't affect performance too much

PR #54, opened 2021-05-19
Add bignum-lib-dependent threshold params

s1na:


bignum_hbls.go
```diff
@@ -0,0 +1,10 @@
+// +build !bignum_pure,!bignum_hol256,!bignum_kilic
```
gballet, line 1: I would just put `!bignum_kilic` here, so that the other two versions (who, incidentally, don't work in go-kzg) and any future version will work here.

tree.go
```diff
@@ -495,7 +495,8 @@ func (n *InternalNode) ComputeCommitment() *bls.G1Point {
 	}
 
 	var commP *bls.G1Point
-	if n.treeConfig.nodeWidth-emptyChildren >= multiExpThreshold {
+	if (n.treeConfig.width == 10 && n.treeConfig.nodeWidth-emptyChildren >= multiExpThreshold10) ||
```
gballet, line 498: I think it would make sense to set `multiExpThreshold` as a member of `TreeConfig`. What do you think?

s1na: Good idea, moved it.

gballet (approved):
LGTM

PR #62, opened 2021-06-22
HashToFr simplified

gballet:
This PR implements a simplified version of `hashToFr` that doesn't rely on the modulo operation. The modulo parameters are therefore removed.

The issue with this, is that the result of the  division mod `2 ** 255` described [here](https://notes.ethereum.org/@vbuterin/verkle_tree_eip) is still larger than the field's order.

### DIFF
--- a/config.go
+++ b/config.go
@@ -26,7 +26,6 @@
 package verkle
 
 import (
-	"math/big"
 	"sync"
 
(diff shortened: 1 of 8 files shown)


gballet:
Superseded by other PRs, closing

PR #63, opened 2021-06-22
tree: apply mod 2^255 in hashToFr

gballet:
The more modest version of #62 : apply the `2^255` modulus before applying that of the field.

Deactivate one test that is no longer valid, to be replaced with the data from the next devnet run.

### DIFF
--- a/tree_test.go
+++ b/tree_test.go
@@ -36,7 +36,6 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/protolambda/go-kzg/bls"
 )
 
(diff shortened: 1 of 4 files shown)


tree_test.go
```diff
@@ -521,55 +520,55 @@ type Account struct {
...
-	tree.Insert(addr2, buf2.Bytes())
-
-	tree.ComputeCommitment()
-
-	block1893Hash := tree.Hash()
-	if !bytes.Equal(block1893Hash[:], common.Hex2Bytes("55938f57d4211b306eb3a1404d4784b2e0a8fdb254f284834b3ccf74791e54ee")) {
-		t.Fatalf("error, got %x != 55938f57d4211b306eb3a1404d4784b2e0a8fdb254f284834b3ccf74791e54ee", block1803Hash)
-	}
-}
+//func TestDevnet0PostMortem(t *testing.T) {
```
s1na, line 523: What's with this test case? I'd suggest either removing it or adding a `t.Skip()` in the beginning if it's not needed temporarily

PR #65, opened 2021-06-23
Implement the last-level node scheme

gballet:
In preparation for the implementation of [the verkle tree eip](https://notes.ethereum.org/@vbuterin/verkle_tree_eip) in geth, change `LeafNode` to be a "last-level node" with `nodeWitdh` values instead of one (basically, replace `LeafNode` with a terminal node).

This PR deactivates the devnet 0 postmortem test, which loses its meaning in this context.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -65,7 +65,7 @@ func ParseNode(serialized []byte, depth, width int) (VerkleNode, error) {
 			return nil, err
 		}
 		var values [][]byte
-		if err := rlp.DecodeBytes(rest, values); err != nil {
+		if err := rlp.DecodeBytes(rest, &values); err != nil {
 			return nil, err
 		}
 		tc := GetTreeConfig(width)
--- a/encoding.go
+++ b/encoding.go
@@ -27,6 +27,7 @@ package verkle
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/rlp"
@@ -69,6 +70,9 @@ func ParseNode(serialized []byte, depth, width int) (VerkleNode, error) {
 			return nil, err
 		}
 		tc := GetTreeConfig(width)
+		if tc.nodeWidth != len(values) {
+			return nil, fmt.Errorf("invalid number of nodes in decoded child expected %d, got %d", tc.nodeWidth, len(values))
+		}
 		ln := &LeafNode{
 			key:        key,
 			values:     values,
--- a/encoding.go
+++ b/encoding.go
@@ -64,11 +64,17 @@ func ParseNode(serialized []byte, depth, width int) (VerkleNode, error) {
 		if err != nil {
 			return nil, err
 		}
-		value, _, err := rlp.SplitString(rest)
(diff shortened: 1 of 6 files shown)


encoding.go
```diff
@@ -64,11 +64,17 @@ func ParseNode(serialized []byte, depth, width int) (VerkleNode, error) {
 		if err != nil {
 			return nil, err
 		}
-		value, _, err := rlp.SplitString(rest)
-		if err != nil {
+		var values [][]byte
+		if err := rlp.DecodeBytes(rest, &values); err != nil {
```
s1na, line 68: Hm, I'm guessing more of this `ParseNode` fn can be re-written to use `DecodeBytes` in future

encoding.go
```diff
@@ -64,11 +64,17 @@ func ParseNode(serialized []byte, depth, width int) (VerkleNode, error) {
...
 			return nil, err
 		}
-		return &LeafNode{key: key, value: value}, nil
+		tc := GetTreeConfig(width)
+		ln := &LeafNode{
+			key:        key,
+			values:     values,
```
s1na, line 74: Maybe add a check to make sure `len(values) == tc.nodeWidth`

tree.go
```diff
@@ -218,15 +221,21 @@ func (n *InternalNode) Insert(key []byte, value []byte) error {
 
 	switch child := n.children[nChild].(type) {
 	case Empty:
-		n.children[nChild] = &LeafNode{key: key, value: value}
+		lastNode := &LeafNode{
+			key:        key,
+			values:     make([][]byte, n.treeConfig.nodeWidth),
+			treeConfig: n.treeConfig,
+		}
+		lastNode.values[key[31]] = value
```
s1na, line 229: `key[31]` should work only for width 8

tree.go
```diff
@@ -351,9 +379,10 @@ func (n *InternalNode) Delete(key []byte) error {
 	case *HashedNode:
 		return errors.New("trying to delete from a hashed subtree")
 	case *LeafNode:
-		if !bytes.Equal(child.key, key) {
+		if !bytes.Equal(child.key[:31], key[:31]) {
 			return errDeleteNonExistent
 		}
+		// overwrite empty if the last level node exists
 		n.children[nChild] = Empty{}
```
s1na, line 386: I feel like you need to call `LeafNode.Delete` here and only replace with `Empty` when all of its values are nil

tree.go
```diff
@@ -575,75 +604,115 @@ func (n *InternalNode) clearCache() {
...
+	return n.values[k[31]], nil
 }
 
 func (n *LeafNode) ComputeCommitment() *bls.G1Point {
-	panic("can't compute the commitment directly")
+	if n.commitment != nil {
+		return n.commitment
+	}
+
+	emptyChildren := 0
```
s1na, line 650: The code below can be turned into a function to be called by both internal and leaf nodes

s1na (changes requested):
I had a comment about `key[31]` only working for width 8. Then I saw you're using that pattern throughout the PR. I guess that means width 8 has been decided on. In that case ignore that comment.

PR #66, opened 2021-06-24
tree: apply mod 2^255 in hashToFr

gballet:
A rewrite of #63. The more modest version of #62 : apply the 2^255 modulus before applying that of the field.

Deactivate one test that is no longer valid, to be replaced with the data from the next devnet run.


s1na (approved):
LGTM

PR #69, opened 2021-06-30
[DONTMERGE] return Frs instead of G1Points in GetCommitmentsAlongPath

gballet:
This was an experiment destined to match more closely the behavior in `compute_commitment_root`, which returns an `int` representing a field element.

This PR has an issue in the calculation of `E` in the verification code: the commitments aren't directly accessible, so `E` can not be calculated by the verifier.

This is a tracking issue to remind me to ensure that the EIP doesn't have this same problem.

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -34,10 +34,10 @@ import (
 	"github.com/protolambda/go-kzg/bls"
 )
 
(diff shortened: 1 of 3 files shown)


gballet:
Superseded by #74, closing

PR #76, opened 2021-07-09
fix: write the trailing bits if the stem isn't byte-aligned

gballet:
Add the trailing bits in the event of a node size of 1024, as the path to the last level will be of 250 bits.


gballet:
That was against the wrong branch, closing

PR #79, opened 2021-07-11
Sync with python impl

gballet:


gballet:
There is still a problem with the proof verification, but the tests now reflect the results of Python.

PR #86, opened 2021-08-13
EIP update: no longer use sha256

gballet:
Update to the latest version of the EIP, in which hash functions are no longer used.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -26,7 +26,6 @@
 package verkle
 
 import (
-	"crypto/sha256"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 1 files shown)


tree.go
```diff
@@ -508,24 +491,15 @@ func (n *InternalNode) ComputeCommitment() *bls.Fr {
 		case *LeafNode:
 			// Store the leaf node hash in the polynomial, even if
 			// the tree is free.
```
gballet, line 493: this comment no longer makes sense and should be removed

gballet:
There's a new proposed model by Dankrad:

image

the best way would be to introduce an `ExtNode` because it effectively splits the `LeafNode` into two, so that the leaves only cover 16 bytes of the actual leaf. The changes are significant, albeit not that complicated, so let's wait for the scheme to settle down before it is implemented.

gballet:
@GottfriedHerold this PR is the basis for replacing BLS with Bandersnatch, it removes the use for hashing. It is not quite up to spec anymore, but have a look nonetheless and let me know what you think.

PR #94, opened 2021-08-31
separate tree structure from commitment calculation

gballet:
Because the commitment algorithm used by verkle trees will evolve, this PR separates the tree structure from the commitment algorith, introducing a new `CommitmentConfig` interface, that provides a `CommitToPoly` function in order to commit to a polynomial.

Co-authored-by: Kevaundray Wedderburn <kevtheappdev@gmail.com>


gballet:
I force-pushed, didn't plan to merge and it merged automatically... something went really wrong here :thinking:

PR #110, opened 2021-10-29
initial attempt at proof of absence

gballet:
This PR won't be merged, it's used as a reference of what should have been done with KZG. The final proof of absence work needs to be performed on the IPA PR #105.

The reason why this work is not complete, is the handling of the  "special case" that disappears in Dankrad-3. Fixing it is not trivial, and the work would be immediately obsolete.


gballet:
Superseded by #111

PR #128, opened 2021-11-10
Open stem at an invalid location to prove absence

### LINKED ISSUE #123: proof of absence: add missing case in which a stem is present
The only type of proof of absence that currently exists, is that where a child is missing in an `InternalNode`. There is also a case in which a `LeafNode` is present, but the stems are different, i.e. `n.key != key`.

gballet:
Fixes #123

### DIFF
--- a/empty.go
+++ b/empty.go
@@ -49,7 +49,7 @@ func (Empty) ComputeCommitment() *Fr {
 	return &FrZero
 }
 
-func (Empty) GetCommitmentsAlongPath([]byte) ([]*Point, []uint8, []*Fr, [][]Fr) {
+func (Empty) GetCommitmentsAlongPath([]byte) *ProofElements {
 	panic("trying to produce a commitment for an empty subtree")
 }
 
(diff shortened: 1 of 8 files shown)


Ssssilverr:
I am very interested in your project "go-verkle". However, I found that the performance have declined with the new pulls. Could you please tell me what caused this? Moreover, is there any solutions to solve this?

gballet:
> I am very interested in your project "go-verkle". However, I found that the performance have declined with the new pulls. Could you please tell me what caused this? Moreover, is there any solutions to solve this?

@Ssssilverr we switched to different cryptographic primitives, that aren't currently optimized. Optimizations are on the way, though.

Ssssilverr:
Hi @gballet ,thank you for your kindly reply.  I just wonder why we need to replace kzg  with ipa.  I did some search and found that, in general  the performance of kzg is as good as IPA...Maybe I’m wrong and I’d love to know your opinion~

PR #135, opened 2021-11-29
Rename LeafNode to ExtNode

gballet:
The historical name of `LeafNode` is no longer describing accurately what this node does: it gathers 256 leaves, sharing the same 31-byte stem. Change the name to something more accurate `ExtNode`, which is short for "extension and suffix tree".


gballet:
Too outdated, will have to be redone

PR #144, opened 2021-12-08
Implement a stateless version of the verkle tree

gballet:
This PR implements a new type of node, meant to create a tree from elements of a proof.


gballet:
Failing tests are probably due to the modification of a global constant...

PR #175, opened 2022-01-14
Add "rust proof format" Deserialization

jwasinger:


proof_ipa.go
```diff
@@ -132,3 +133,66 @@ func SerializeProof(proof *Proof) ([]byte, error) {
...
+	}
+	poaStems = make([][]byte, numPoaStems)
+	for i := 0; i < int(numPoaStems); i++ {
+		var poaStem [31]byte
+		if err := binary.Read(reader, binary.LittleEndian, &poaStem); err != nil {
+			return nil, err
+		}
+
+		poaStems[i] = poaStem[:]
+	}
```
gballet, line 158: This allocate a buffer and copies the data into it, when you already have direct access to the source data.

```suggestion
	// verify that there is enough data to read
	if numPoaStems * 31 + 4 <= len(proofSerialized) {
		return nil, errors.New("truncated proof")
	}
	poaStems = make([][]byte, numPoaStems)
	for i := 0; i < int(numPoaStems); i++ {
		poaStems[i] = poaStem[4+31*i:4+31*i+31]
		reader.Seek(31, io.SeekStart)
	}
```

proof_ipa.go
```diff
@@ -132,3 +133,66 @@ func SerializeProof(proof *Proof) ([]byte, error) {
...
+			return nil, err
+		}
+		if err := commitment.Unmarshal(commitmentBytes); err != nil {
+			return nil, err
+		}
+
+		commitments[i] = &commitment
+	}
+
+	// TODO submit PR to go-ipa to make this return an error if it fails to Read
```
gballet, line 189: do me a favor and create an issue for this, instead of leaving the TODO here. Admittedly, all the TODOs will eventually be fixed, but that might take some time.

proof_test.go
```diff
@@ -310,3 +310,44 @@ func TestProofSerializationWithAbsentStem(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(serialized) == 0 {
+		t.Fatal("zero-length serialized proof payload")
+	}
+
+	deserialized, err := DeserializeProof(serialized)
+	if err != nil {
+		t.Fatal("could not deserialize verkle proof")
```
gballet, line 345:
```suggestion
		t.Fatalf("could not deserialize verkle proof: %v", err)
```

proof_ipa.go
jwasinger: Okay.  If we go with this option, then the caller must make sure that the source data is not used after they call `DeserializeProof(sourceData)`.  It has potential to create some annoying bugs down the line.

But I'm fine to make it this way if you think the reduction in allocations is worth it here.

proof_ipa.go
jwasinger: https://github.com/crate-crypto/go-ipa/issues/12

proof_ipa.go
gballet: It's only a problem if the buffer is reused to deserialize another proof afterwards, it's unlikely to happen but I guess you are right. I made a note of this in issue #176 and we will revisit it when the go-ethereum part of the code is more stable, and we know that this won't happen.

gballet (approved):
LGTM

PR #185, opened 2022-02-03
fix endianness issues in commitment calculations

gballet:
The endianness in commitments was wrong, this PR fixes it.

### DIFF
--- a/ipa.go
+++ b/ipa.go
@@ -44,7 +44,7 @@ func CopyPoint(dst, src *Point) {
 
 func toFr(fr *Fr, p *Point) {
 	bytes := p.Bytes()
-	fr.SetBytes(bytes[:])
+	fr.SetBytesLE(bytes[:])
 }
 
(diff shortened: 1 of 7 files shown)


gballet:
failing tests: looks like the endianness problem is also present in the proof, which is to be expected looking at the multipoint part.

ipa.go
```diff
@@ -44,18 +44,19 @@ func CopyPoint(dst, src *Point) {
...
 
 func from32(fr *Fr, data [32]byte) {
 	fr.SetBytes(data[:])
 }
 
 func FromLEBytes(fr *Fr, data []byte) {
+	var aligned [32]byte
 	for i := range data {
-		data[i], data[len(data)-1-i] = data[len(data)-1-i], data[i]
+		aligned[31-i] = data[i]
```
gballet, line 57: this loop is probably unnecessary when calling `SetBytesLE`

PR #193, opened 2022-02-23
Fix case in Insertion where a HashedNode resolves into a LeafNode

jwasinger:
this should result in a branch (new `InternalNode`s created)  if the LeafNode key is not the exact key we are inserting on.


gballet (approved):
LGTM

PR #211, opened 2022-05-10
Helper functions for flushing nodes to disk during conversion

gballet:
In order not to run out of memory, be a tad more aggressive when flushing a node. This means that deeper nodes will have to be reloaded from the DB, but the memory should remain below 16GB.

This is one of 3 methods that are considered. The other two will receive their own PRs.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -443,6 +443,13 @@ func (n *InternalNode) FlushStem(stem []byte, flush NodeFlushFn) {
 	child := n.children[nChild]
 	switch child := child.(type) {
 	case *InternalNode:
+		if n.depth == 2 {
+			child.ComputeCommitment()
+			child.Flush(flush)
+			n.children[nChild] = child.toHashedNode()
+			return
+		}
+
 		child.FlushStem(stem, flush)
 	case *LeafNode:
 		if child.commitment == nil {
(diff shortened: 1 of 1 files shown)


gballet:
First 8 hours of running this PR, the RAM usage seems to tend towards an asymptote (that was expected) at around 8GB (tbc).

The required disk space grows linearly (also expected).

8h

5 million accounts have been covered so far, which means that if the data can be extrapolated (it can not, most of the big contracts are still ahead) it should take **13 days** to do the translation with this method. The final size should then be of about ~120GB.

gballet:
The memory growth towards an asymptote seems to continue, and the state continues to grow linearly. The memory seem that it will stabilize around 8GB and the state will grow towards ~250GB and the translation will take 20 days. As said before, the amount of data that is processed tends to vary and in my experience, the "end" of the address space is more empty. Still, a worrisome trend.

24h

Note that this approach has several flaws:
 - internal nodes are stored by commitment. When they get flushed to disk, then reloaded from disk and inserted into, their commitment will change. The consequence is that the older version of the node will remain in the database and take up space
 - the "3rd level" limit is not very smart and will flush items regardless of the memory available: if the translating node has more memory, that memory isn't used to limit the amount of flushes/recovery from disk.
 - values can not be inserted in order, which would allow for a very efficient algorithm.

PR #214, opened 2022-05-16
add a lock to run flushing and insertion in parallel

gballet:
TODO:

 * [x] write a test for a leafnode split upon node resolution
 * [x] fix proof verification tests
 * [x] insert multiple values at the same stem (see #217)
 * [ ] dedup insertion code


gballet:
reminder: don't merge before I wrote a test for a leafnode split upon node resolution

gballet (commented):
Dedup-ing the code is quite messy and it would make the PR more bug-prone at this stage. I'll dedeup after the commitment update on insert.

Issue #136, opened 2021-11-29
Update commitment on insert


gballet:
Currently, the commitment is calculated once and for all. Insertion ends up being faster, however this is a problem for stateless execution, as data is missing from the tree.

A previous implementation existed, which was KZG-based, and would have to be rewritten to work with IPA. The idea is to introduce a new internal node type, that maintains a `map[byte]VerkleNode` instead of an array.

gballet:
The purpose of the stateful tree is to perform the conversion as fast as possible. Recomputing a commitment when many inserts are happening in the same branch is going to make things slower. Better reserve diff-commitment insertion for stateless tree.

In case we need to go back on this decision, a fuzzed implementation is available at #219

PR #219, opened 2022-05-19
diff commitment calculation in stateful Insert

gballet:


gballet:
See #136 for an explanation of why this PR is closed.

Issue #229, opened 2022-06-04
Stateful delta insert/delete


gballet:
As explained in #136, the main goal of `Insert` and `Delete` for `InternalNode` and `LeafNode` is to be used for the conversion.

There is still a need for a version of these methods that does delta insertion/deletion, so that the whole commitment is not needed. This should speedup the stateful block executions. Moist of the code is already present in #219, it just needs to be reworked a bit.

gballet:
done with the CoW stuff, closing

PR #231, opened 2022-06-07
add marker to differentiate the hash of a leaf from of an internal node

gballet:
Necessary for #229

This is useful because it makes it possible to diff-insert into `LeafNodes` without actually loading them in the tree. This has two consequences:

* No need to store the `LeafNodes` in the DB - they are considered write-only
* Save time during conversion because a vast majority of the nodes aren't saved to the database


gballet:
The current approach is not optimal, size-wise. since `HashedNode`s are still stored individually instead of being serialized along with their parents nodes. This would improve read performance because de-serializing the child node requires as many database jumps as there are hashed leaf nodes.

gballet:
We no longer reference children by hash, so this is no longer useful. Closing.

PR #241, opened 2022-07-06
GetProofItems for StatelessNode

gballet:
This is necessary for the verification of a serialized proof, as some elements are found by running `GetProofItems`

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -278,7 +278,7 @@ func TreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) {
 			si.stem = stems[stemIndex]
 			si.values = map[byte][]byte{}
 			for i, k := range proof.Keys {
-				if !bytes.Equal(k[:31], si.stem) && proof.Values[i] != nil {
+				if bytes.Equal(k[:31], si.stem) && proof.Values[i] != nil {
 					si.values[k[31]] = proof.Values[i]
 					si.has_c1 = si.has_c1 || (k[31] < 128)
 					si.has_c2 = si.has_c2 || (k[31] >= 128)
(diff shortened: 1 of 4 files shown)


proof_ipa.go
```diff
@@ -278,7 +278,7 @@ func TreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) {
 			si.stem = stems[stemIndex]
 			si.values = map[byte][]byte{}
 			for i, k := range proof.Keys {
-				if !bytes.Equal(k[:31], si.stem) && proof.Values[i] != nil {
```
gballet, line 281: weird that I never caught this before, check if this happens on the testnet

tree.go
```diff
@@ -137,7 +137,9 @@ func (pe *ProofElements) Merge(other *ProofElements) {
 		pe.Cis = append(pe.Cis, ci)
 		pe.Zis = append(pe.Zis, other.Zis[i])
 		pe.Yis = append(pe.Yis, other.Yis[i])
-		pe.Fis = append(pe.Fis, other.Fis[i])
+		if pe.Fis != nil {
+			pe.Fis = append(pe.Fis, other.Fis[i])
+		}
```
gballet, line 142: Add a comment that `Fis` will be `nil` in stateless mode

proof_ipa.go
gballet: because on the testnet, the output of that function is never used.

stateless.go
```diff
@@ -387,8 +389,164 @@ func (n *StatelessNode) ComputeCommitment() *Point {
...
+		}
+
+		// Loop over again, collecting the children's proof elements
+		// This is because the order is breadth-first.
+		for _, group := range groups {
+			childIdx := offset2key(group[0], n.depth)
+
+			// Special case of a proof of absence: no children
+			// commitment, as the value is 0.
+			if n.children[childIdx] == nil {
```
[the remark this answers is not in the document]
gballet: that was just a poor translation of the stateful version, the fix is correct

PR #252, opened 2022-07-22
remove the delete method from VerkleNode

gballet:
Verkle trees do not support deletion. The value should be overwritten with 0s instead. As new code paths get added with e.g. differential insertion, reducing the maintenance scope by removing methods that are not necessary make sense.


gballet:
I am not sure I should merge this, because a distinction has been made on the geth side, between the trie interface, concerned with the format (`StateTrie`) and its backed, format-agnostic implementation (`trie.Trie`). The same distinction can be made with `VerkleTrie` (the format-concerned trie) and its format-agnostic backend (`gballet/go-verkle.VerkleNode`).

`VerkleNode` can support deletion, as long as `VerkleTrie` does not call it. That's why I wanted to remove it from the trie in the first place: make sure that no one accidentally calls `(VerkleNode).Delete`. But it might be useful in the future. Keeping it, however, means that it is likely that this will bitrot.

gballet:
This PR no longer makes sense, closing.

PR #253, opened 2022-07-22
differential insertion on hashed leaf nodes

gballet:
Second experiment for addressing #229 (see #231) : the signature of `Insert` is changed so that when values are inserted, and if a node is a hashed leaf node (i.e. a `HashedNode` with `stem != nil`), then no value is actually inserted and its commitment is computed from the delta between the old value and the new one.

Although its semantics haven't changed yet, `LeafNode` is expected, in the future, to only be used to initialize an extension-and-leaf-node, and never for an update. The end goal is to only use it into `InsertStemOrdered` and discontinue both `InsertOrdered` and `InsertStem` (if possible, it might be difficult to achieve  because of an out-of-order conversion process).

With this method, `LeafNodes` are never directly stored into the DB: values are read from the snapshot, and provide all the necessary information to update the tree without storing the serialized list of values. This makes the whole tree smaller both in RAM and on disk. It will also reduce IO during the conversion process (pending a modification to embed the hashed node in its parent `InternalNode`) and root commitment calculations (no full polynomial evaluation at each node)

TODO:
 * [x] also update the commitment of its parents
 * [x] fix `TestInsertIntoHashedNode`
 * [ ] also implement it for `StatelessNode`
 * [ ] rebase once the hashed leaf nodes are stored directly into their parent `InternalNode` for reduced I/O.
 * [ ] add a fuzzer against a `LeafNode` tree

Out of scope:
 * Removal of `InsertStem` and `InsertOrdered`
 * diff-insertion of multiple values into the same leaf

### DIFF
--- a/benchs/main.go
+++ b/benchs/main.go
@@ -47,7 +47,7 @@ func benchmarkInsertInExisting() {
 		for i := 0; i < 5; i++ {
 			root := verkle.New()
 			for _, k := range keys {
-				if err := root.Insert(k, value, nil); err != nil {
+				if err := root.Insert(k, nil, value, nil); err != nil {
 					panic(err)
 				}
 			}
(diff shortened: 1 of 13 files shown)


gballet:
This has bitrotted for a long time, and most of its ideas are now implemented as part of other PRs. Closing.

PR #259, opened 2022-08-26
internal node diff-insert

gballet:
Simplified version of #253 that only diff-updates the internal nodes.


gballet:
superseded by #261

PR #272, opened 2022-09-20
Rename ComputeCommitment to Commitment

gballet:
This PR is the first step in implementing batch inserts. It cleans up the API by separating the commitment getter from the function compuring it (code sold separately).

**TODO**

 - [ ] Rewrite `InsertStemOrdered` to support diff-insert, so it no longer relies on `ComputeCommitment` to get its value


gballet:
Done in another PR, closing.

PR #279, opened 2022-10-27
revert diff-update in insertordered + add non-update insert

gballet:
This PR:

 1, Reverts the changes that made `InsetOrdered` to update its commitments on insert, as they made no sense: `InsertOrdered` is meant to be used in the conversion process, and updating the commitment at each insert will cause more computations than necessary, especially in the upper nodes;
 2. Introduces a new `InsertNoCommUpdate` method on `InternalNode`, that has the same behavior as `Insert` before the update-on-insert PR was merged.

The goal of this PR is to speed up conversion, it should not affect performance of the block replay process.

I am not yet sure that this PR will be merged, as I am considering a more thorough rewrite of the API: set the update behavior at node creation time, remove `InsertNoCommUpdate`, and keep a list of touched values so that only these are recomputed by `Commit`, as called by `IntermediateRoot`.

### DIFF
--- a/empty.go
+++ b/empty.go
@@ -29,7 +29,11 @@ import "errors"
 
 type Empty struct{}
 
-func (Empty) Insert([]byte, []byte, NodeResolverFn) error {
+func (n Empty) Insert(k []byte, v []byte, r NodeResolverFn) error {
+	return n.insert(k, v, r, true)
+}
+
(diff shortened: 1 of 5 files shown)


gballet:
superseded by #292. Closing.

PR #280, opened 2022-11-03
copy-on-write commitments, take 1

gballet:
This is a first attempt at creating copy-on-write commitments, so that only the contribution for the values that were created are recomputed. This PR has a few issues, and will probably be rewritten and broken down into smaller pieces.


gballet:
Superseded by #288, closing

PR #288, opened 2022-11-08
Implement CoW for stateless internal nodes

gballet:
Implement copy-on-write of commitments for `StatelessNode` as well.

This effectively breaks stateless leaf nodes, which isn't a problem because it is not used in geth at the moment, and a rewrite of the child node management is necessary before geth can execute statelessly again.


gballet:
Already merged into master.

Issue #270, opened 2022-09-15
perform batch insertions


gballet:
With `InsertAtStem`, multiple values are inserted into the same extension-and-suffix-tree. This is nice, because the cost of updating the commitments along the path are shared.

Insertions should be batched so that the same technique can be used at all levels, and particularly at the root.

gballet:
Another approach has been taken in #303 with the copy-on-write approach: the commitment won't be recomputed immediately, but a list of changed locations is kept. This way, only actually modified values are updated when calling `Commit`, and all at once.

PR #291, opened 2022-11-08
bench: group to field

jsign:
```
BenchmarkGroupToField/single-16                   793231              1506 ns/op
BenchmarkGroupToField/multiple/1-16               539619              1862 ns/op              1862 ns/value          144 B/op          6 allocs/op
BenchmarkGroupToField/multiple/2-16               520746              2251 ns/op              1125 ns/value          336 B/op          8 allocs/op
BenchmarkGroupToField/multiple/4-16               343665              3501 ns/op               875.2 ns/value        736 B/op         12 allocs/op
BenchmarkGroupToField/multiple/8-16               178898              6015 ns/op               751.8 ns/value       1530 B/op         17 allocs/op
BenchmarkGroupToField/multiple/16-16              116660             10384 ns/op               649.0 ns/value       3124 B/op         25 allocs/op
BenchmarkGroupToField/multiple/32-16               67484             17460 ns/op               545.6 ns/value       6312 B/op         39 allocs/op
BenchmarkGroupToField/multiple/64-16               35632             35094 ns/op               548.4 ns/value      12713 B/op         68 allocs/op
BenchmarkGroupToField/multiple/128-16              18733             65418 ns/op               511.1 ns/value      25435 B/op        114 allocs/op
BenchmarkGroupToField/multiple/256-16               8793            137116 ns/op               535.6 ns/value      50922 B/op        210 allocs/op
```

PR #292, opened 2022-11-10
Use CoW in node insert + code simplification

gballet:
This PR is a major refactor that is needed to introduce various optimizations for the block replay:

Optimization:

 * Use a "copy-on-write" approach in `InternalNode` : the commitment of a node is updated from only the commitments of its children that were written to with `Insert` or `Delete`. That computation is intended to be performed once, at the end of a block creation/verification.
 * Use `MapToScalarField` to speed up some stuff, although this is already covered by #291 so it won't be integrated here.

Refactoring:

  * `StatelessNode` and `InternalNode` no longer use diff-insert, as multiple children will typically be updated during block execution
  * `LeafNode` uses diff-insert, as insertion into this type of nodes typically happen only once per block
  * `LeafNodeCommit` therefore no longer perform any calculation
  * `InternalNode.Commit` on the other hand, will only recompute the commitment contribution from the nodes that were written to.
  * Simplify the code by merging `Insert` and `InsertStem`, as well as `InsertOrdered` and `InsertStemOrdered`.

This PR shaves roughly 20% of the execution time, compared to `master`.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -36,8 +36,6 @@ var (
 	errNotSupportedInStateless = errors.New("not implemented in stateless")
 	errInsertIntoOtherStem     = errors.New("insert splits a stem where it should not happen")
 	errStatelessAndStatefulMix = errors.New("a stateless node should not be found in a stateful tree")
-	errUnknownNodeType         = errors.New("unknown node type")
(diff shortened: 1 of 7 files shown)


gballet:
Fixes #270

gballet:
Rebased in #303

PR #296, opened 2022-11-12
update GetConfig() error handling

bshiluk:
Use panic() instead of printing errors. Remove Get Config() error return value and update all references.


gballet (approved):
LGTM, I have pushed to your branch to fix the deepsource warning, I will merge when it passes.

PR #301, opened 2022-11-22
set of changes needed to re-enable the verkle tests in go-ethereum

gballet:
This only fixes the verkle tests, not all geth tests, but it's enough to produce sample blocks


gballet:
This is already covered by #302, closing

PR #302, opened 2022-11-23
Use serialized commitments in HashedNode

gballet:
Until now, when de-serializing an `InternalNode`, the commitments of its 256 children would also be deserialized in order to instantiate them as `HashedNode`. This is wasteful, as this deserialization operaiton is expensive and, when resolving a `HashedNode`, the commitment is serialized again in order to get the serialized child node from the database.

This PR modifies `HashedNode` to only contain the serialized commitment.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -47,27 +47,9 @@ func ParseNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
 	if len(serialized) < 64 {
 		return nil, serializedPayloadTooShort
 	}
-
 	switch serialized[0] {
 	case leafRLPType:
-		var values [NodeWidth][]byte
-		offset := 64
-		for i := 0; i < NodeWidth; i++ {
-			if bit(serialized[32:64], i) {
-				if offset+32 > len(serialized) {
-					return nil, fmt.Errorf("verkle payload is too short, need at least %d and only have %d, payload = %x (%w)", offset+32, len(serialized), serialized, serializedPayloadTooShort)
-				}
-				values[i] = serialized[offset : offset+32]
-				offset += 32
-			}
-		}
-		if NodeWidth != len(values) {
-			return nil, fmt.Errorf("invalid number of nodes in decoded child expected %d, got %d", NodeWidth, len(values))
-		}
-		ln := NewLeafNode(serialized[1:32], values[:])
-		ln.setDepth(depth)
-		ln.Commit()
-		return ln, nil
+		return parseLeafNode(serialized, depth)
 	case internalRLPType:
 		return CreateInternalNode(serialized[1:33], serialized[33:], depth, comm)
 	default:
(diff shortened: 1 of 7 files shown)


stateless.go
```diff
@@ -813,10 +806,7 @@ func (n *StatelessNode) toInternalNode() *InternalNode {
 		if child, ok := n.children[byte(i)]; ok {
 			internal.children[i] = child
 		} else if serialized, ok := n.unresolved[byte(i)]; ok {
-			hashed := &HashedNode{hash: new(Fr), commitment: new(Point)}
-			hashed.commitment.SetBytesTrusted(serialized)
-			toFr(hashed.hash, hashed.commitment)
-			internal.children[byte(i)] = hashed
+			internal.children[byte(i)] = &HashedNode{serialized}
```
gballet, line 809: this whole function can be deleted because it was only used in one test, that has been modified

tree.go
```diff
@@ -394,7 +394,7 @@ func (n *InternalNode) InsertStem(stem []byte, node VerkleNode, resolver NodeRes
 		if resolver == nil {
 			return errInsertIntoHash
 		}
-		hash := child.commitment.Bytes()
+		hash := child.commitment
```
gballet, line 397: this change also needs to be applied line 293

gballet (commented):
looks good so far, now I get a couple panics when I run the tests, due to the fact that `ParseNode` no longer returns a `StatelessNode` and some type assertions need to change from `(*StatelessNode)` to `.(*InternalNode)` in `tree_test.go`.

For the test failing in `stateless_test.go`, I guess it makes sense to create a function called `ParseStatelessNode` that retains the old behavior, and is called any time one really wants to deserialize into a `StatelessNode`.

hashednode.go
```diff
@@ -55,14 +54,16 @@ func (n *HashedNode) Commit() *Point {
 	if n.commitment == nil {
 		panic("nil commitment")
 	}
-	return n.commitment
+	c := new(Point)
+	c.SetBytes(n.commitment)
```
gballet, line 58: Could you use `SetBytesTrusted` here? For increased performance.

stateless.go
```diff
@@ -252,22 +252,19 @@ func (n *StatelessNode) Insert(key []byte, value []byte, resolver NodeResolverFn
...
 			// fallthrough to hash resolution
 		}
 
 		// If the child is a hash, the node needs to be resolved
 		// before there is an insert into it.
 		if h, ok := n.children[nChild].(*HashedNode); ok {
-			comm := h.Commitment().Bytes()
-			serialized, err := resolver(comm[:])
+			comm := h.commitment
+			serialized, err := resolver(comm)
```
gballet, line 263: you can use `serialized, err := resolver(n.commitment)`

stateless.go
```diff
@@ -252,22 +252,19 @@ func (n *StatelessNode) Insert(key []byte, value []byte, resolver NodeResolverFn
...
 			if err != nil {
 				return err
 			}
-			node, err := ParseNode(serialized, n.depth+1, comm[:])
+			node, err := ParseNode(serialized, n.depth+1, comm)
```
gballet, line 267: `ParseStatelessNode` should be used here: if you are using a 'stateless' tree, then resolving a node should yield stateless nodes.

So `ParseStatelessNode` would need to be amended to also handle leaf values (i.e. just have the old behavior). See comment above.

encoding.go
```diff
@@ -69,12 +69,22 @@ func ParseNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
...
+func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
+	if len(serialized) < 64 {
+		return nil, serializedPayloadTooShort
+	}
+	if serialized[0] == internalRLPType {
+		return deserializeIntoStateless(serialized[1:33], serialized[33:], depth, comm)
+	}
+	return nil, ErrInvalidNodeEncoding
+}
+
```
gballet, line 87: That's nice, but is only able to desserialize a statless, internal node. That's a problem In the case of the stateless insert, for example, because hashnode resolution in stalessnode insert should return stateless internal nodes. So the best way would be to make a function that has the old behavior, including being able to return `LeafNode` when deserializing a leaf.

I understand it's code duplication, but it's possible to use a helper function to reduce the amount of duplicated code, and hopefully `StatelessNode` is going to disappear eventually.

gballet (commented):
Looks nice, I have a couple extra comments and we should be good to go.

PR #309, opened 2022-12-01
hashednode: cache commitment resolution

### LINKED ISSUE #307: double commitment deserialization for HashedNode in InsertStem
With #303, `n.cowChild` will call `Commitment` who will in turn call `Commit`. This might be a problem in the case of a `HashedNode`, since it will be deserialized twice: once when calling `cowChild` and once when resolving it (e.g. in `CreateInternalNode`, called by `ParseNode`).

One idea would be to modify the signature of `NodeResolverFn` to accept an already-resolved commitment.

bshiluk:
Cache resolved commitment in `HashedNode{resolution:...}` to prevent the potential for duplicate deserialization. Fixes #307.

### DIFF
--- a/hashednode.go
+++ b/hashednode.go
@@ -31,8 +31,8 @@ import (
 )
 
 type HashedNode struct {
-	commitment []byte
-	resolution *Point
+	commitment  []byte
+	cachedPoint *Point
 }
 
 func (*HashedNode) Insert([]byte, []byte, NodeResolverFn) error {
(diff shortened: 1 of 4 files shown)


hashednode.go
```diff
@@ -32,6 +32,7 @@ import (
 
 type HashedNode struct {
 	commitment []byte
+	resolution *Point
```
gballet, line 35: I would prefer to call it `cachedPoint` or something that explicitly states it's a point

gballet (commented):
Good so far. I have a minor nitpick, and I want to test this change, I'll let you know when it's done.

gballet (approved):

gballet:
image
20s improvement in overall performance, always nice to have

PR #317, opened 2022-12-21
serialize leaf node commitments for cheaper deserialization

bshiluk:


gballet (approved):
LGTM. I still have to run it on my test machine, and then I will merge it.

gballet:
The gain is only 10s, but still nice to have.
profile046

I need to update CI before merging this, otherwise the tests will fail.

PR #318, opened 2023-01-11
make a nil value json-encode to an empty string in KeyValuePair

gballet:
Uses `gencodec` to ensure that a `KeyValuePair` with a `nil` `Value` will turn into a JSON representation with an empty string as its `value` member.

This is used for the verkle-enabled lighthouse prototype, until the point when the SSZ format is fully defined.


gballet:
This is no longer required as we are moving to a proof format closer to the final one.

PR #320, opened 2023-01-19
allow parallel node flushing for top layer

jsign:
This PR improves the flushing performance by parallelizing the work in a controlled manner.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -29,6 +29,8 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


jsign:
@gballet, more PR cleaning here. I'll close it since this parallelization hasn't shown to be justified the overhead at the runtime level compared to the parallelized work.

Maybe a good idea to reconsider later if we have another kind of workload for the replay benchmark, but today there's no good argument to consider it.

PR #322, opened 2023-01-19
Batch toFr in proof calculation & go-ipa update

jsign:
This PR batches multiple improvements:
- Update to the latest go-ipa which has an improvement regarding memory allocations.
- In proof calculations, leverage the batched-Fr calculations which improves performance.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,6 @@ module github.com/gballet/go-verkle
 
 go 1.18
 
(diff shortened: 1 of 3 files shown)


jsign:
@gballet, this PR is ready to be considered again... it has an important improvement for proof generation reg inverse batching. Also, updating ot the latest `go-ipa@master` to not fall behind much (and it has some extra mem alloc optimization from some time ago). Not sure if you want to measure again just to know.

gballet (approved):
LGTM

PR #323, opened 2023-01-31
serialization: manage uncompressed point serialization for trusted sources

jsign:
This PR is an experimental change to avoid persisting compressed EC points in the resolver. 
This will allow parsing nodes to use less CPU since no `Y` coordinate should be calculated from `X`.

While doing this change, I switched multiple places with "magic numbers" with more explicit calculations and constants.

This PR is using a [custom go-ipa version](https://github.com/crate-crypto/go-ipa/pull/33).

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -44,7 +44,7 @@ func bit(bitlist []byte, nr int) bool {
 var serializedPayloadTooShort = errors.New("verkle payload is too short")
 
 func ParseNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
-	if len(serialized) < 1+31+SerializedAffinePointSize {
+	if len(serialized) < 1+StemSize+SerializedAffinePointSize {
 		return nil, serializedPayloadTooShort
 	}
 	switch serialized[0] {
@@ -58,7 +58,7 @@ func ParseNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
 }
 
 func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
-	if len(serialized) < 1+31+SerializedAffinePointSize {
+	if len(serialized) < 1+StemSize+SerializedAffinePointSize {
 		return nil, serializedPayloadTooShort
 	}
 	switch serialized[0] {
@@ -73,9 +73,9 @@ func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode,
 
 func parseLeafNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
 	var values [NodeWidth][]byte
-	offset := 1 + 31 + 32 + 2*SerializedAffinePointSize
+	offset := 1 + StemSize + 32 + 2*SerializedAffinePointSize
 	for i := 0; i < NodeWidth; i++ {
-		if bit(serialized[1+31:1+31+32], i) {
+		if bit(serialized[1+StemSize:1+StemSize+32], i) {
 			if offset+32 > len(serialized) {
 				return nil, fmt.Errorf("verkle payload is too short, need at least %d and only have %d, payload = %x (%w)", offset+32, len(serialized), serialized, serializedPayloadTooShort)
 			}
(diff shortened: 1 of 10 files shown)


encoding.go
```diff
@@ -44,21 +44,21 @@ func bit(bitlist []byte, nr int) bool {
 var serializedPayloadTooShort = errors.New("verkle payload is too short")
 
 func ParseNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
-	if len(serialized) < 64 {
+	if len(serialized) < 1+StemSize+SerializedAffinePointSize {
```
jsign, line 47: Transforming magic numbers into more obvious constant calculations. This wasn't trivial now that the size of the affine point changed it's size.

encoding.go
```diff
@@ -44,21 +44,21 @@ func bit(bitlist []byte, nr int) bool {
...
 		return nil, serializedPayloadTooShort
 	}
 	switch serialized[0] {
 	case leafRLPType:
 		return parseLeafNode(serialized, depth, comm)
 	case internalRLPType:
-		return CreateInternalNode(serialized[1:33], serialized[33:], depth, comm)
+		return CreateInternalNode(serialized[1:1+NodeWidth/8], serialized[1+NodeWidth/8:], depth, comm)
```
jsign, line 54: `NodeWith/8` is a clearer calculation to signal the bitfield size.

encoding.go
```diff
@@ -86,12 +86,12 @@ func parseLeafNode(serialized []byte, depth byte, comm []byte) (VerkleNode, erro
 	if NodeWidth != len(values) {
```
jsign, line 86: All these are doing two things at once:
- Avoiding more magic numbers
- Indirectly the `SerializedAffinePointSize` has been switched from 32 to 64, since now both points are part of the serialization.

encoding_test.go
```diff
@@ -18,7 +18,8 @@ func TestLeafStemLength(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(ser) != 128 {
+	// RLP_type + Stem + bitset + 2-commitments
+	if len(ser) != 1+StemSize+NodeWidth/8+2*SerializedAffinePointSize {
```
jsign, line 22: Went a bit further with a comment to explain this that wasn't trivial.

go.mod
```diff
@@ -5,3 +5,5 @@ go 1.18
 require github.com/crate-crypto/go-ipa v0.0.0-20221111143132-9aa5d42120bc
 
 require golang.org/x/sys v0.0.0-20220919091848-fb04ddd9f9c8 // indirect
+
+replace github.com/crate-crypto/go-ipa => github.com/jsign/go-ipa v0.0.0-20230131214742-21080f1193c2
```
jsign, line 9: @gballet, I had to do this since the `go-ipa-` branch in https://github.com/crate-crypto/go-ipa/pull/33 isn't part of the `go-ipa` repo.

If you're going to test this in the benchmark, you'll have to add this `replace` line in the `go-ethereum` `go.mod` file since `replaces` aren't "inherited" in dependencies.

ipa.go
```diff
@@ -31,10 +31,13 @@ import (
...
 
 type (
-	Fr    = fr.Element
-	Point = banderwagon.Element
+	Fr                    = fr.Element
+	Point                 = banderwagon.Element
+	SerializedAffinePoint = banderwagon.SerializedAffinePoint
 )
 
+const SerializedAffinePointSize = banderwagon.SerializedAffinePointSize
```
jsign, line 39: New type and constant from go-ipa that are pulled here to be more explicit and type safe.

proof_ipa.go
```diff
@@ -138,7 +138,7 @@ func SerializeProof(proof *Proof) ([]byte, []KeyValuePair, error) {
```
jsign, line 138: Remember that for proofs, we still use compressed serialization.

gballet: proofs and commitments computation alike. With this approach, I get a different root commitment, so something must impact commitment calculation, when it shouldn't.

jsign: All tests, proof gen, and verification still pass.
One of the tests is checking compatibility with Rust and is still passing.

This sounds more like a `go-ethereum` call to the changed semantics `Bytes()` as we chatted; we can look at it.

jsign:
@gballet, I'm closing this PR, and I'll open a separate one with these changes plus other optimizations in a separate one.
I'll be addressing all your comments in this PR in the new version, so comments/feedback aren't lost.

PR #324, opened 2023-02-03
update (Des|S)erializeProof to the electra proof  format

gballet:
This is a first attempt at using the new [electra proof format](https://github.com/ethereum/consensus-specs/pull/3230).

It still uses the intermediate `Proof` structure, which was used to deserialize the binary proof format. The final version of this code should remove this intermediate step entirely. But this change is already complex enough to warrant an intermediate step. Nonetheless, it this new proof format would simplify `TreeFromProof` as there is no longer a need to go over multiple keys with the same stem.


gballet (commented):
it's ready to go, but in order to be merged, the block replay code has to be successfully rebased on top of `beverly-hills-head` otherwise a lot of things are going to break and it will become a maintenance nightmare.

PR #326, opened 2023-02-14
fix: issue with proof of absence stems

gballet:


gballet:
weird issue: a block replay from rust-verkle will fail iff the other tests are running. And in that case, one of the commitments corresponding to a C1 subtree is missing (although its children are included)

gballet:
this isn't a beverly hills issue at all, it's a go-verkle issue

gballet:
as suggested by @jsign this is an outdated PR for code that has been significantly rewritten since. closing.

PR #327, opened 2023-02-15
Switch to uncompressed form serialization

jsign:
This PR is an improved version of a [previous iteration](https://github.com/gballet/go-verkle/pull/323) of this work, plus some additional performance improvements.

This PR is building on [top of another go-ipa PR](https://github.com/crate-crypto/go-ipa/pull/36). That PR is similar to a previous iteration. Still, it has changed semantics as suggested to keep `Bytes()` to have the same "compressed" interpretation we had before and create new methods for _uncompressed_ form.

Moreover, this PR will be used in another PR in go-ethereum soon, which finishes a combo of go-ethereum+go-verkle+go-ipa changes to improve performance.

I'll add comments, but still, keep this PR as a draft until we validate the results.

[All this work is part of improving the replay benchmark in `go-ethereum`](https://github.com/gballet/go-ethereum/pull/164),

### DIFF
--- a/config.go
+++ b/config.go
@@ -32,10 +32,11 @@ import (
 const (
 	NodeWidth         = 256
 	NodeBitWidth byte = 8
+	StemSize          = 31
 )
 
 func equalPaths(key1, key2 []byte) bool {
-	return bytes.Equal(key1[:31], key2[:31])
+	return bytes.Equal(key1[:StemSize], key2[:StemSize])
 }
 
 // offset2key extracts the n bits of a key that correspond to the
--- a/config.go
+++ b/config.go
@@ -30,9 +30,10 @@ import (
 )
 
 const (
-	NodeWidth         = 256
-	NodeBitWidth byte = 8
-	StemSize          = 31
+	LeafValueSize      = 32
+	NodeWidth          = 256
+	NodeBitWidth  byte = 8
+	StemSize           = 31
 )
 
 func equalPaths(key1, key2 []byte) bool {
--- a/encoding.go
+++ b/encoding.go
@@ -101,20 +101,19 @@ func ParseStatelessNode(serialized []byte, depth byte, comm SerializedPointCompr
 }
 
 func parseLeafNode(serialized []byte, depth byte, comm SerializedPointCompressed) (VerkleNode, error) {
-	// Create a slice of children values making sure they're 32 bytes long (zero-padded).
-	var paddedChildrenValues [NodeWidth][]byte
 	bitlist := serialized[leafBitlistOffset : leafBitlistOffset+bitlistSize]
+	var values [NodeWidth][]byte
 	offset := leafChildrenOffset
 	for i := 0; i < NodeWidth; i++ {
 		if bit(bitlist, i) {
 			if offset+LeafValueSize > len(serialized) {
 				return nil, fmt.Errorf("verkle payload is too short, need at least %d and only have %d, payload = %x (%w)", offset+32, len(serialized), serialized, errSerializedPayloadTooShort)
 			}
-			paddedChildrenValues[i] = serialized[offset : offset+LeafValueSize]
+			values[i] = serialized[offset : offset+LeafValueSize]
 			offset += LeafValueSize
 		}
 	}
-	ln := NewLeafNodeWithNoComms(serialized[leafSteamOffset:leafSteamOffset+StemSize], paddedChildrenValues[:])
+	ln := NewLeafNodeWithNoComms(serialized[leafSteamOffset:leafSteamOffset+StemSize], values[:])
 	ln.setDepth(depth)
 	ln.c1 = new(Point)
 	ln.c1.SetBytesUncompressed(serialized[leafC1CommitmentOffset:leafC1CommitmentOffset+SerializedPointSize], true)
--- a/encoding.go
+++ b/encoding.go
@@ -41,25 +41,25 @@ func bit(bitlist []byte, nr int) bool {
 	return bitlist[nr/8]&mask[nr%8] != 0
 }
 
-var serializedPayloadTooShort = errors.New("verkle payload is too short")
+var errSerializedPayloadTooShort = errors.New("verkle payload is too short")
 
-func ParseNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
-	if len(serialized) < 64 {
-		return nil, serializedPayloadTooShort
+func ParseNode(serialized []byte, depth byte, comm SerializedPointCompressed) (VerkleNode, error) {
+	if len(serialized) < 1+StemSize+SerializedPointSize {
+		return nil, errSerializedPayloadTooShort
 	}
 	switch serialized[0] {
 	case leafRLPType:
 		return parseLeafNode(serialized, depth, comm)
 	case internalRLPType:
-		return CreateInternalNode(serialized[1:33], serialized[33:], depth, comm)
+		return CreateInternalNode(serialized[1:1+NodeWidth/8], serialized[1+NodeWidth/8:], depth, comm)
 	default:
 		return nil, ErrInvalidNodeEncoding
 	}
 }
 
-func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
-	if len(serialized) < 64 {
-		return nil, serializedPayloadTooShort
+func ParseStatelessNode(serialized []byte, depth byte, comm SerializedPointCompressed) (VerkleNode, error) {
+	if len(serialized) < 1+StemSize+SerializedPointSize {
+		return nil, errSerializedPayloadTooShort
 	}
 	switch serialized[0] {
 	case leafRLPType:
@@ -71,13 +71,13 @@ func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode,
 	}
 }
 
-func parseLeafNode(serialized []byte, depth byte, comm []byte) (VerkleNode, error) {
+func parseLeafNode(serialized []byte, depth byte, comm SerializedPointCompressed) (VerkleNode, error) {
 	var values [NodeWidth][]byte
-	offset := 128
+	offset := 1 + StemSize + 32 + 2*SerializedPointSize
 	for i := 0; i < NodeWidth; i++ {
-		if bit(serialized[32:64], i) {
+		if bit(serialized[1+StemSize:1+StemSize+32], i) {
 			if offset+32 > len(serialized) {
-				return nil, fmt.Errorf("verkle payload is too short, need at least %d and only have %d, payload = %x (%w)", offset+32, len(serialized), serialized, serializedPayloadTooShort)
+				return nil, fmt.Errorf("verkle payload is too short, need at least %d and only have %d, payload = %x (%w)", offset+32, len(serialized), serialized, errSerializedPayloadTooShort)
 			}
 			values[i] = serialized[offset : offset+32]
 			offset += 32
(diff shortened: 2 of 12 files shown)


config.go
```diff
@@ -30,12 +30,14 @@ import (
 )
 
 const (
-	NodeWidth         = 256
-	NodeBitWidth byte = 8
+	LeafValueSize      = 32
+	NodeWidth          = 256
+	NodeBitWidth  byte = 8
+	StemSize           = 31
```
jsign, line 36: Same with this.

encoding.go
```diff
@@ -71,74 +100,76 @@ func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode,
...
 	}
 	n.commitment = new(Point)
-	n.commitment.SetBytesTrusted(comm)
+	n.commitment.SetBytes(comm, true)
 	return n, nil
 }
 
 func indicesFromBitlist(bitlist []byte) []int {
-	indices := make([]int, 0)
+	indices := make([]int, 0, 32)
```
jsign, line 165: This obvious optimization saved a ton of allocations and GC pressure, which looks obvious in hindsight.
We can only have up to 32 indices, so we do a single allocation with that capacity.

This `indicesFromBitlist(...)` was called _a lot_ of times in the replay benchmark from `go-ethereum`.

encoding.go
```diff
@@ -71,74 +100,76 @@ func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode,
...
 	for i, b := range bitlist {
 		if b == 0 {
 			continue
 		}
 		// the bitmap is little-endian, inside a big-endian byte list
 		for j := 0; j < 8; j++ {
-			if b&mask[j%8] != 0 {
+			if b&mask[j] != 0 {
```
jsign, line 172: Not needed since `j = [0, 8)`. Maybe the compiler was smart enough to optimize this, but let's no rely on that.

hashednode.go
```diff
@@ -31,14 +31,37 @@ import (
 )
```
jsign, line 31: OK, important changes are coming here. This was an important optimization for the replay benchmark.

hashednode.go
```diff
@@ -31,14 +31,37 @@ import (
...
 
 type HashedNode struct {
-	commitment  []byte
+	commitment  SerializedPointCompressed
```
jsign, line 34: Here I just formalize the type to avoid confusion if `[]byte` is compressed or uncompressed.
`commitment` is the compressed serialization of the point. Same semantics as before, but being clear here.

We can't assume we have this value always available. The serialization of points now are in uncompressed form. Transforming that to compressed is CPU costly.

`commitment` will be calculated _if_ needed and in the fastest possible way depending on the context. 
More about this to be explained in next comments. The TL;DR: think of it as a "cached" value.

hashednode.go
```diff
@@ -31,14 +31,37 @@ import (
...
+	serialized  SerializedPoint
```
jsign, line 35: `serialized` is a new field, which is exactly the serialized representation that the DB has.
Now this is our source of truth for `HashedNode`. We _always_ have value in this field.

As in, when we want to create a _HashedNode_, we create it with the serialized bytes we got from the database. No CPU work, just storing.

hashednode.go
```diff
@@ -31,14 +31,37 @@ import (
...
 	cachedPoint *Point
 }
 
+func NewHashedNode(buf SerializedPoint) *HashedNode {
+	var h HashedNode
+	h.serialized = buf
+
+	return &h
+}
```
jsign, line 44: Before, we didn't have a constructor, and that felt like the wild-west on how this _HashedNode_ was created.
I wanted to centralize the logic in a clear place, and that place is here.

Also, as explained before, it is clear that the way to create a _HashedNode_ is with the serialized bytes (i.e: uncompressed form of the point).

hashednode.go
```diff
@@ -31,14 +31,37 @@ import (
...
+
 func (*HashedNode) Insert([]byte, []byte, NodeResolverFn) error {
 	return errInsertIntoHash
 }
 
+func (h *HashedNode) BytesUncompressed() []byte {
+	return h.serialized
+}
```
jsign, line 52: New method that it's used to have the serialized version of this _HashedNode_. This is trivial in this case, since our _HashedNode_ representation is literally holding the serialized bytes we got from the DB, so we simply return that.

hashednode.go
```diff
@@ -31,14 +31,37 @@ import (
...
+func (h *HashedNode) BytesCompressed() []byte {
+	if h.commitment == nil {
+		if h.cachedPoint != nil {
+			h.commitment = h.cachedPoint.Bytes()
+			return h.commitment
+		}
+		h.commitment = h.serialized.ToCompressed()
+	}
+	return h.commitment
+}
```
jsign, line 63: OK, so here is the work we wanted to delay as much as possible.
Now _HashedNode_ is asked to return the compressed version of the point.

Our source of truth always available is `serialized` (i.e: uncompressed form of the point), but we take some strategy here:
1. If `h.commitment != nil` (i.e: we already calculated this before and have it cached), use it obviously.
2. If that isn't the case, we've never calculated `h.commitment`. Do we have `h.cachedPoint` cached? If that's the case, let's get the compressed form by getting it from the in-memory point to `Bytes()`. That's faster than our last resource (next point)
3. What if we don't have that `h.cachedPoint`? We don't have any other option than do the work of transforming from our uncompressed serialized representation (i.e: `h.serialized`) to the compressed serialization form. We do this with a new method I created in go-ipa PR `.ToCompressed()`. 

That `ToCompressed()` method is somewhat costly... if we can optimize that, there's another big opportunity to improve performance. This is the topic I mentioned in our chat that I pinged @kevaundray; we can explore that whenever he can think about it.

The TL;DR of all this comment is: 
- Delay `h.commitment` calculation as much as we can
- Use whatever opportunity we have regarding other cached values to do this computation as fast as we can.

This change was very important for the replay `go-ethereum` gains.

hashednode.go
```diff
@@ -52,19 +75,19 @@ func (*HashedNode) Get([]byte, NodeResolverFn) ([]byte, error) {
 }
 
 func (n *HashedNode) Commit() *Point {
-	if n.commitment == nil {
-		panic("nil commitment")
+	if n.serialized == nil {
+		panic("invalid hashed node")
 	}
```
jsign, line 80: Remember: source of truth always available is `serialized` not `commitment`.

hashednode.go
```diff
@@ -78,11 +101,15 @@ func (*HashedNode) Serialize() ([]byte, error) {
...
-	c := &HashedNode{commitment: make([]byte, len(n.commitment))}
-	copy(c.commitment, n.commitment)
+
+	c := &HashedNode{
+		serialized: make([]byte, len(n.serialized)),
+	}
+	copy(c.serialized, n.serialized)
+
 	return c
 }
```
jsign, line 114: That's it... not that wild change, but very convoluted to figure out via benchmarks from go-ethereum that this path was very hot.

go.mod
```diff
@@ -5,3 +5,5 @@ go 1.18
 require github.com/crate-crypto/go-ipa v0.0.0-20221111143132-9aa5d42120bc
 
 require golang.org/x/sys v0.0.0-20220919091848-fb04ddd9f9c8 // indirect
+
+replace github.com/crate-crypto/go-ipa => github.com/jsign/go-ipa v0.0.0-20230215213932-c52c461f8ea7
```
jsign, line 9: We have to remove this whenever we merge the go-ipa PR. For now, I'm replacing it directly.

stateless.go
```diff
@@ -78,9 +78,7 @@ func NewStateless() *StatelessNode {
 }
```
jsign, line 78: I accommodated things in this file to make it compile, but it won't work.
I didn't care much since @gballet mentioned that we'll kill all this anyway.

stateless_test.go
```diff
@@ -472,6 +472,10 @@ func TestStatelessInsertIntoHash(t *testing.T) {
 // This test checks that a serialized node will be deserialized before
 // being inserted into, during leaf insertion.
 func TestStatelessInsertIntoSerialized(t *testing.T) {
+	// Skipped since this test isn't compatible anymore with the new HashedNode version,
+	// and we're planning to deprecate "stateless trees" anyway.
+	t.SkipNow()
```
jsign, line 477: Skipped this test and the next one for the same reasons. Making this impl work isn't easy with the new changes, but we're planning to kill it anyway.

I'm not sure when that is planned, but now it looks like it's a good idea to do it asap.

tree.go
```diff
@@ -537,13 +536,13 @@ func (n *InternalNode) Get(k []byte, getter NodeResolverFn) ([]byte, error) {
...
 		if err != nil {
 			return nil, err
 		}
 
 		// deserialize the payload and set it as the child
-		c, err := ParseNode(payload, n.depth+1, child.commitment)
+		c, err := ParseNode(payload, n.depth+1, child.BytesCompressed())
 		if err != nil {
 			return nil, err
 		}
```
jsign, line 548: All the **above** changes in this file are using the new API of `HashedNode`. Nothing new in the mechanics, just avoiding private fields and being clear about what we're getting.

tree.go
```diff
@@ -569,13 +568,6 @@ func (n *InternalNode) Commitment() *Point {
 }
 
 func (n *InternalNode) Commit() *Point {
-	// If we're at the first or second layer, we do the Commit() in parallel
-	// leveraging available cores.
-	// If we're in further layers, we do single-goroutine Commit() since the top
-	// two layers already created enough goroutines doing downstream work.
-	if n.depth <= 1 {
-		return n.commitParallel()
-	}
```
jsign, line 578: Now that I did other optimizations, still looking at benchmarks I noticed this parallelization started to hurt.

Things were much faster now, and we're doing less allocations (so less GC pressure). At this point I noticed this was putting a lot of pressure on the scheduler of goroutines and doing some extra heap escaping that affected performance.

This isn't a mere "let's simplify now that we're faster"... this was literally slowing things down by a 2-digit %.
The faster you go, the more these "parallelization overhead" start to count.

I think this is great... we're faster using a single core!

tree.go
```diff
@@ -622,93 +614,6 @@ var frPool = sync.Pool{
 	},
 }
 
-func (n *InternalNode) commitParallel() *Point {
```
jsign, line 625: :wave:

tree.go
```diff
@@ -820,24 +725,33 @@ func (n *InternalNode) Serialize() ([]byte, error) {
 			}
 		}
 	}
-	children := make([]byte, 0, (len(commitments)+nhashed)*32)
 
-	bytecomms := banderwagon.ElementsToBytes(commitments)
+	ret := make([]byte, nodeTypeSize+bitlistSize+(len(commitments)+nhashed)*SerializedPointSize)
```
jsign, line 729: OK, some optimizations happening in this method which makes things lighter for the CPU and produce less allocation & GC pressure.

Before in L823 (deleted), we created the `children` slice and later did some "appending" to produce the final serialized result.

Now we do something smarter. I create the `ret` byte slice to be returned here, and everything will be calculated directly in this slice.

tree.go
```diff
@@ -820,24 +725,33 @@ func (n *InternalNode) Serialize() ([]byte, error) {
...
+
+	// We create a children slice from ret ready to start appending children without allocations.
+	children := ret[internalNodeChildrenOffset:internalNodeChildrenOffset]
```
jsign, line 732: We still use the `children` slice. But now `children` is a slice of the final returned slice. Kind of an internal-view, so no copies/append is needed later.

In `ret` creation, we already allocated exactly the amount of bytes as capacity that we need to make `append`ing to `children` to be allocation free.

tree.go
```diff
@@ -820,24 +725,33 @@ func (n *InternalNode) Serialize() ([]byte, error) {
...
 				consumed++
 			}
 		}
 	}
 
-	return append(append([]byte{internalRLPType}, bitlist[:]...), children...), nil
+	// Store in ret the serialized result
+	ret[nodeTypeOffset] = internalRLPType
+	copy(ret[internalBitlistOffset:], bitlist[:])
+	// Note that children were already appended in ret through the children slice.
```
jsign, line 752: Since `children` was already written in `ret` (via `children` internal slice), we only set the node type, and copy the bitlist which is 32 bytes (quick).

tree.go
```diff
@@ -1254,22 +1168,36 @@ func (n *LeafNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]byte
 	return pe, esses, poass
 }
 
+// Serialize serializes a LeafNode.
+// The format is: <nodeType><stem><bitlist><c1comm><c2comm><children...>
```
jsign, line 1172: Helping future readers.

tree.go
```diff
@@ -1254,22 +1168,36 @@ func (n *LeafNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]byte
...
 func (n *LeafNode) Serialize() ([]byte, error) {
-	var bitlist [32]byte
-	children := make([]byte, 0, NodeWidth*32)
+	// Empty value in LeafNode used for padding.
+	var emptyValue [LeafValueSize]byte
```
jsign, line 1175: I'm avoiding here what we did previously in L1265 (`padding := make([]byte, 32-len(v))`). 

The compiler might not be smart enough to figure out that we don't need to allocate there, and we're always using it for the same purpose.

So now we create this `emptyValue`, which is a "empty leaf value" (i.e: zero 32 byte), and re-use this when padding values below. Also, I create it as an array, not a slice. Just to be sure this small array will live on the stack.

tree.go
```diff
@@ -1254,22 +1168,36 @@ func (n *LeafNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]byte
...
+	children := make([]byte, 0, NodeWidth*LeafValueSize)
+	var bitlist [bitlistSize]byte
 	for i, v := range n.values {
 		if v != nil {
 			setBit(bitlist[:], i)
 			children = append(children, v...)
-			if len(v) < 32 {
-				padding := make([]byte, 32-len(v))
+			if padding := emptyValue[:32-len(v)]; len(padding) != 0 {
 				children = append(children, padding...)
```
jsign, line 1185: Every padding reuses the `emptyValue` and that's fine. `append` is copying always zero-bytes.

tree.go
```diff
@@ -1254,22 +1168,36 @@ func (n *LeafNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]byte
...
+
+	// Create the serialization.
+	baseSize := nodeTypeSize + StemSize + bitlistSize + 2*SerializedPointSize
+	result := make([]byte, baseSize, baseSize+4*32) // Extra pre-allocated capacity for 4 values.
+	result[0] = leafRLPType
+	copy(result[leafSteamOffset:], n.stem[:StemSize])
+	copy(result[leafBitlistOffset:], bitlist[:])
+	copy(result[leafC1CommitmentOffset:], n.c1.BytesUncompressed())
+	copy(result[leafC2CommitmentOffset:], n.c2.BytesUncompressed())
+	result = append(result, children...)
```
jsign, line 1198: This is pulling another optimization in a [pending PR](https://github.com/gballet/go-verkle/pull/322). It wasn't entirely central to the gains, but it avoids allocations, and that's good. Also, it makes clearer the result construction compared to nested appends.

encoding.go
```diff
@@ -71,74 +100,76 @@ func ParseStatelessNode(serialized []byte, depth byte, comm []byte) (VerkleNode,
...
+	if len(raw)/SerializedPointSize != len(indices) {
 		return nil, ErrInvalidNodeEncoding
 	}
+
+	freelist := make([]HashedNode, len(indices))
 	for i, index := range indices {
-		n.children[index] = &HashedNode{commitment: raw[i*32 : (i+1)*32]}
+		freelist[i].serialized = raw[i*SerializedPointSize : (i+1)*SerializedPointSize]
+		n.children[index] = &freelist[i]
 	}
```
jsign, line 158: This innocent change saved billions of allocations (I mean, _literally_), which greatly impacted the replay benchmark (i.e: double digit % improvement).
What happened here is that each `&HashedNode{..}` call would make a heap allocation, since that's a pointer stored as a `VerkleNode` node.

That's unfortunate, since we're creating _a lot_ of `HashedNodes`, and that means a lot of `mallocgc` calls plus a lot of GC pressure.

So, what I did is simple. We already know here how many `HashedNode`s we'll need. I make a single slice allocation of that size, and then initialize from there. That means only 1 alloc vs worst-case 256. Regarding "freeing" semantics, all is fine, since all _HashNodes_ are children of the same parent, so all of them will be "dropped" at the same time liberating the slice memory.

All this is is another example of why it's useful to see pprofs of e2e benchmarks; not the usual thing you'll imagine it will save a billion allocations and save so many seconds of CPU.

jsign:
Some update about this draft PR:
- I've to rebase it to `master` to clean up things and leave a clean "proposal" to consider uncompressed form serialization.
- This PR is `on-hold` since it isn't entirely clear yet we want to pay the cost of storing more info despite the CPU savings. That's not viable today that we aren't doing "in-memory pruning", but maybe later the disk-IO hit won't be that big and can be reconsidered.

TL;DR for now marking this as `on-hold`.

ws38483746:
good!

PR #328, opened 2023-02-21
serialization refactor & optimizations

jsign:
This PR contains all the changes from https://github.com/gballet/go-verkle/pull/327, but keeps the compressed (existing) serialization.

I'll write the replay results for my machine and the AWS VM here.

### DIFF
--- a/config.go
+++ b/config.go
@@ -30,12 +30,14 @@ import (
 )
 
 const (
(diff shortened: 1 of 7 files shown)


jsign:
@gballet, maybe you want to double-check yourself, but looks good. (Gains without disk IO changes)

PR #329, opened 2023-02-22
Add two new serializations formats for internal nodes

jsign:
This PR introduces two new serializations for internal nodes, so we have three ones now:
- "Bitmap": is the only one we have had up to now; we use a 32-byte bitmap to mark the index of the serialized children.
- "Full": doesn't have a 32-byte bitmap or any overhead since it's assumed **all** 256 children are non-zero.
- "Offset": doesn't use a 32-byte bitlist. But encodes each child as `<1-byte-offset><children-0><1-byte-offset><children-1>...` (e.g: `<12><32-byte-compressed-point><24><32-byte-compressed-point>` means that the first child is index 12, and second index 24). 1-byte for the offset is enough since the index is at most 255.

The strategy to decide which serialization to use is the following:
- If all the children are non-zero, use "Full" serialization.
- If the number of non-zero children is <32, use "Offset" serialization.
- If none of the above holds, use the usual "Bitmap" serialization.

There's a different idea to try, too: doing the "inverse inclusion" (i.e: mark which children **aren't** set).

This PR doesn't touch leaf node serialization.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -77,10 +77,17 @@ func ParseNode(serializedNode []byte, depth byte, comm SerializedPointCompressed
 	switch serializedNode[0] {
 	case leafRLPType:
 		return parseLeafNode(serializedNode, depth, comm)
-	case internalRLPType:
+	case internalRLPTypeBitmap:
 		bitlist := serializedNode[internalBitlistOffset : internalBitlistOffset+bitlistSize]
 		children := serializedNode[internalNodeChildrenOffset:]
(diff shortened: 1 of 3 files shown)


jsign:
This PR still sounds like a good idea to save bytes in internal-nodes depending on the number of non-zero children.

For now, we have other more important performance-related PRs in the pipeline, so putting this on pause for a bit. We can come back to this in some weeks.

jsign:
With PBSS we already did other serialization changes.

PR #331, opened 2023-02-26
Fix proof element JSON serialization for compatibility with lighthouse

gballet:


gballet (commented):
There was a successful devnet launched, merging.

PR #332, opened 2023-02-27
tree: batch Point->Fr transformations per level when updating commitments

jsign:
This PR changes how we `Commit()` to the tree to batch more field inversions. 

Previously, we had a recursive algorithm to update the commitments in the tree. On each node, we batched the point->Fr mapping of children. This was fine, but there was an optimization opportunity to do bigger batches.

The only dependency to update the commitments at depth N is that we have updated the commitments at level N+1. This means that we could batch the point->Fr mapping per level, which is "bigger" than "per node".

This avoids some amount of CPU-seconds in our replay benchmark.

More details about the code change in the PR comments.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -29,7 +29,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"sync"
 
 	"github.com/crate-crypto/go-ipa/banderwagon"
 )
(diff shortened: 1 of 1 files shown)


tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
 func (n *InternalNode) Commit() *Point {
-	if len(n.cow) != 0 {
-		polyp := frPool.Get().(*[]Fr)
-		poly := *polyp
-		defer func() {
-			for i := 0; i < NodeWidth; i++ {
-				poly[i] = Fr{}
+	if len(n.cow) == 0 {
+		return n.commitment
+	}
```
jsign, line 581: If cow is empty, then we can do an early return since all our subtree was untouched.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
+
+	internalNodeLevels := make([][]*InternalNode, StemSize)
+	n.fillLevels(internalNodeLevels)
```
jsign, line 584: We fill in `internalNodeLevels` all the **touched** nodes by level. i.e: `internalNodeLevels[N]` contains a list of `InternalNode`s at depth N.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
+func (n *InternalNode) fillLevels(levels [][]*InternalNode) {
+	levels[int(n.depth)] = append(levels[int(n.depth)], n)
+	for _, child := range n.children {
+		if childInternalNode, ok := child.(*InternalNode); ok && len(childInternalNode.cow) > 0 {
+			childInternalNode.fillLevels(levels)
+		}
+	}
+}
+
 func (n *InternalNode) Commit() *Point {
```
jsign, line 578: I think the Github diffs might be confusing, may be better to see the code change locally or with split-views.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
+		return n.commitment
+	}
+
+	internalNodeLevels := make([][]*InternalNode, StemSize)
+	n.fillLevels(internalNodeLevels)
+
+	points := make([]*Point, 0, 1024)
+	cowIndexes := make([]int, 0, 1024)
+	poly := make([]Fr, NodeWidth)
+	for level := len(internalNodeLevels) - 1; level >= 0; level-- {
```
jsign, line 589: The big-picture is that we'll do the commitment updating per level, starting from "the bottom" and going up.

What we're doing is similar to the previous logic to update commitments in a single node:
1. We do a previous step to collect all the previous/post touched commitments from the level.
2. We do a single `toFrMultiple` in the level.
3. Now that we have the Frs, we do the diff-updating taking the points from there.

I'll explain each step below.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
+		nodes := internalNodeLevels[level]
+		if len(nodes) == 0 {
+			continue
+		}
+		points = points[:0]
+		cowIndexes = cowIndexes[:0]
```
jsign, line 595: `points` and `cowIndexes` (to be explained below) are slices defined outside of the L589 loop. This was just to avoid allocating per level, and reusing slices cleaning them up per level.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
-		var i int
-		b := make([]byte, len(n.cow))
-		points := make([]*Point, 2*len(n.cow))
-		for idx, comm := range n.cow {
-			emptyChildren--
-			points[2*i] = comm
-			points[2*i+1] = n.children[idx].Commit()
-			b[i] = idx
-			i++
 		}
```
jsign, line 606: This is step 1. We iterate in all the nodes in the level, and collects in `points` the previous (cow) and new commitments that we'll have to Fr to do the diff updating later.

`cowIndexes` is an auxiliary slice that indicates that the i-th element corresponds to the idx-th element in the node cow. This is needed since doing `range node.cow` has random order (since it's a map). So we can't infer `idx` from the iteration order!

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
 			frs[i] = &Fr{}
 		}
-		toFrMultiple(frs, points)
-		for i := 0; i < len(points)/2; i++ {
-			poly[b[i]].Sub(frs[2*i+1], frs[2*i])
-		}
 
-		n.cow = nil
+		// Do a single batch calculation for all the points in this level.
+		toFrMultiple(frs, points)
```
jsign, line 615: This is step 2. We do a single `toFrMultiple` for this level. Most of the CPU-heavy work happens here.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
 
-		n.commitment.Add(n.commitment, cfg.CommitToPoly(poly, emptyChildren))
-		return n.commitment
+		// We calculate the difference between each (new commitment - old commitment) pair, and store it
+		// in the same slice to avoid allocations.
+		for i := 0; i < len(frs); i += 2 {
+			frs[i/2].Sub(frs[i+1], frs[i])
+		}
+		// Now `frs` have half of the elements, and these are the Frs differences to update commitments.
+		frs = frs[:len(frs)/2]
```
jsign, line 623: I calculate the difference between the correspondings Frs while leveraging the same `frs` slice to avoid allocating another slice. Then I reslice it to give it the appropriate length.

Now in `frs` we have the "differences" between old and new Frs.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
+			}
+			for i := 0; i < len(node.cow); i++ {
+				poly[cowIndexes[cowIndex]] = *frs[frsIdx]
+				frsIdx++
+				cowIndex++
+			}
+			node.cow = nil
+			node.commitment.Add(node.commitment, cfg.CommitToPoly(poly, 0))
+		}
 	}
```
jsign, line 640: This is step 3. Here we do some careful work to take from the "big" `frs` slice (which contains "diffs" from all the nodes in this level) to update the commitments as corresponds. Saying it differently, we can iterate again in `nodes`, taking chunks of `frs` that correspond to each node and doing the updating.

In L629-L631 we're just doing a cleaning of `poly` that we reuse in each level to avoid allocations.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
...
-
 	return n.commitment
 }
 
-var frPool = sync.Pool{
-	New: func() any {
-		ret := make([]Fr, NodeWidth)
-		return &ret
-	},
-}
```
jsign, line 615: Since I was quite aggressive in avoiding allocations in the new `Commit()` implementation, we can remove this pool. I've double-checked the allocations in the replay benchmark, and the new `Commit()` isn't moving the needle now so no need to do pooling.

tree.go
```diff
@@ -567,53 +566,81 @@ func (n *InternalNode) Commitment() *Point {
 	return n.commitment
 }
 
+func (n *InternalNode) fillLevels(levels [][]*InternalNode) {
+	levels[int(n.depth)] = append(levels[int(n.depth)], n)
+	for _, child := range n.children {
```
gballet, line 571: I'm missing the point here: why not iterate over `n.cow` instead of all the children?

gballet (commented):
LGTM. I left a comment and I'm running it on my test machine, when all of that is addressed, it'll be merged.

tree.go
jsign: There's no point actually, it will save some cycles, but doubt it will make a difference. In any case, is a good idea.
I changed it to use `cow` now (last commit).

gballet (approved):
LGTM

PR #333, opened 2023-03-03
Enable batch serialization

jsign:
This PR creates a new `(*InternalNode).BatchSerialize(...)` API. 

The motivation is to continue pushing more batching to avoid CPU costs doing field inverses. While serializing touched nodes to be saved in the database, we must transform in-memory projective points to their compressed form. That process involves transforming from projective to affine, which requires division (one per **commitment** inside the node, so there can be plenty).

The idea is to switch from a recursive style as we had in `Flush(...)` to this new `BatchSerialize(...)`, which is iterative. We do a "deep scan" of all non-hashed (i.e: touched) nodes internal Points (i.e: commitments c1, c2, internal node children, etc), and we do a _single_ `ElementsToBytes(...)` call which already allowed batching projective->compressed transformations.

Review notes:
- I won't enable this PR for review until https://github.com/gballet/go-verkle/pull/332 is merged. This PR is sitting on top of that one, and I'd like to rebase to `master` to leave as changes only the relevant code.


tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+
+// BatchSerialize is an optimized serialization API when multiple VerkleNodes serializations are required, and all are
+// available in memory.
+func (n *InternalNode) BatchSerialize() ([]SerializedNode, error) {
+	// Commit to the node to update all the nodes commitments.
+	n.Commit()
+
+	// Collect all nodes that we need to serialize.
+	nodes := make([]VerkleNode, 0, 1024)
+	nodes = n.collectNonHashedNodes(nodes)
```
jsign, line 1284: We collect in `nodes` all touched nodes from the tree (i.e: non-hash nodes; that is: present internal and leaf nodes).

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+
+	// We collect all the *Point so we can batch invert the work the projective->affine transformation.
```
jsign, line 1286: In `pointsToCompress` we range over `nodes` and try to collect here all the points we need to serialize.

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+			count := 0
+			for _, c := range n.children {
+				_, isInternalNode := c.(*InternalNode)
+				_, isLeafNode := c.(*LeafNode)
+				if isInternalNode || isLeafNode {
+					pointsToCompress = append(pointsToCompress, c.Commitment())
+					count++
+				}
+			}
+			pointToCompressPerNodeCount[i] = 1 + count // Internal node commitment + all non-hashed children.
```
jsign, line 1302: For an internal node, we add in `pointsToCompress` all the non-hashed children commitments plus its own commitment.

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+		case *LeafNode:
+			pointsToCompress = append(pointsToCompress, n.commitment, n.c1, n.c2)
+			pointToCompressPerNodeCount[i] = 3 // Leaf node commitment + c1 + c2.
```
jsign, line 1305: For leaf nodes, we collect in `pointsToCompress` both `c1` and `c2`, plus its own commitment.

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+	// Commit to the node to update all the nodes commitments.
+	n.Commit()
+
+	// Collect all nodes that we need to serialize.
+	nodes := make([]VerkleNode, 0, 1024)
+	nodes = n.collectNonHashedNodes(nodes)
+
+	// We collect all the *Point so we can batch invert the work the projective->affine transformation.
+	pointsToCompress := make([]*Point, 0, len(nodes)*NodeWidth+len(nodes))
+	pointToCompressPerNodeCount := make([]int, len(nodes))
```
jsign, line 1288: This is a helper slice to track how many points were collected for compression per `node` item. It just helps after collecting to keep track of "from whom" each result is for.

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+			}
+			pointToCompressPerNodeCount[i] = 1 + count // Internal node commitment + all non-hashed children.
+		case *LeafNode:
+			pointsToCompress = append(pointsToCompress, n.commitment, n.c1, n.c2)
+			pointToCompressPerNodeCount[i] = 3 // Leaf node commitment + c1 + c2.
+		}
+	}
+
+	// Now we do the all transformations in a single-shot.
+	compressedPoints := banderwagon.ElementsToBytes(pointsToCompress)
```
jsign, line 1310: Now that we have in `pointsToCompress` all we need, let's do the compressed representation in a single shot _for everything_.

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+			c2Bytes := compressedPoints[1]
+			sn := SerializedNode{
+				Node:            n,
+				CommitmentBytes: compressedPoints[0],
+				SerializedBytes: n.serializeWithCompressedCommitments(c1Bytes, c2Bytes),
+			}
+			ret = append(ret, sn)
+		}
+		compressedPoints = compressedPoints[pointToCompressPerNodeCount[i]:]
+	}
```
jsign, line 1336: ~TL;DR: collect the results, reconstruct things in right order, and return.

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+	result := make([]byte, baseSize, baseSize+4*32) // Extra pre-allocated capacity for 4 values.
+	result[0] = leafRLPType
+	copy(result[leafSteamOffset:], n.stem[:StemSize])
+	copy(result[leafBitlistOffset:], bitlist[:])
+	copy(result[leafC1CommitmentOffset:], c1Bytes[:])
+	copy(result[leafC2CommitmentOffset:], c2Bytes[:])
+	result = append(result, children...)
+
+	return result
+}
```
jsign, line 1420: A final comment about this PR.
I kept the existing `Flush(...)` API despite we [moved away from using it in go-etheruem](https://github.com/gballet/go-ethereum/pull/172).

Maybe it makes sense to keep it if for other uses cases it might be useful, but [I'd say probably isn't](https://github.com/gballet/go-ethereum/pull/172#discussion_r1124345678). In any case, if we later decide to remove that API, I think we might need to do a decent refactor in tests. (That's fine, but if that's the direction we're going let's plan for that in a separate PR)

tree.go
gballet: won't that make things redundant then? If the child's commitment is included, then it has been modified. If the child is an internal node, then it will also have its children's commitments as well as its own.

tree.go
```diff
@@ -1293,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+			} else {
+				children = append(children, compressedChildren[consumed][:]...)
+				consumed++
+			}
+		}
+	}
+
+	// Store in ret the serialized result
+	ret[nodeTypeOffset] = internalRLPType
+	copy(ret[internalBitlistOffset:], bitlist[:])
```
gballet, line 1386: why not use `ret[internalBitlistOffset:]` directly instead of copying the bitlist? Exactly what you do with `children` and `ret`.

tree.go
gballet: Yes, let's slowly convert the tests over time, but as long as it is used in tests, `Flush` is useful. It also needs to be replaced in other branches, like the new replay/convert branches I hope to release in the next few days.

tree.go
```diff
@@ -1293,3 +1266,155 @@ func ToDot(root VerkleNode) string {
...
+		}
+		compressedPoints = compressedPoints[pointToCompressPerNodeCount[i]:]
+	}
+
+	return ret, nil
+}
+
+func (n *InternalNode) collectNonHashedNodes(list []VerkleNode) []VerkleNode {
+	list = append(list, n)
+	for _, child := range n.children {
```
gballet, line 1343: see my 2nd question in the general review summary: would it be possible to use `n.cow` here?

gballet (commented):
Nice work1 A couple questions/suggestions:

1. I think there is some redundancy since, in order to compute the commitment to the parent, the child's commitment already has to be serialized. Caching the serialization done during the commit and reusing it during the parent's serialization could prove to be quite a gain.
2. By default, a `HashedNode` doesn't have a deserialized commitment. Only the children that are accessed/resolved are turned into points. But there could be nodes that are deserialized just for reading, and they would have to be serialized again, even though its serialization was already known through the `HashedNode`'s `commitment` field, which is a `[]byte`. This can potentially save a lot of computations when reading code/storage. And line 1343, you would iterate over `n.cow` instead of `n.children`.

I haven't run it yet, I'll publish my test results tomorrow.

tree.go
```diff
@@ -1266,3 +1266,155 @@ func ToDot(root VerkleNode) string {
 	root.Commit()
 	return fmt.Sprintf("digraph D {\n%s}", root.toDot("", ""))
 }
+
+type SerializedNode struct {
+	Node            VerkleNode
+	CommitmentBytes [32]byte
+	SerializedBytes []byte
+}
```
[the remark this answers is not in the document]
jsign: Sure, that sounds like a good idea.  I'll add add comments. 
Maybe soon, after som things stabilize, we can do a full scan and improve on the documentation front.

tree.go
jsign: Yeah, looks a bit confusing, I'll improve the comment.

tree.go
jsign: > Yes, let's slowly convert the tests over time, but as long as it is used in tests, Flush is useful. 
:+1: 

>  It also needs to be replaced in other branches, like the new replay/convert branches I hope to release in the next few days.

Yeah, I did that already in https://github.com/gballet/go-ethereum/pull/172 that you'll review next.

tree.go
jsign: Uhm, unfortunately it looks that it won't work in that case and I think I know the reason.

The Verkle Trie is called `Commit()` before the `BatchSerialize()` call in go-ethereum. It first does [trie.Hash()](https://github.com/gballet/go-ethereum/blob/verkle/replay-change-with-tree-group-tryupdate/core/state/statedb.go#L979), which does `tire.root.Commit().Bytes()`... so unfortunately at the start of `BatchSerialize()` we already lost the `cows`.

This for me opens an even bigger question that probably you'll know better. That `IntermediateRoot` is committing _per transaction and not block-wise? If that's the case, the new batching that we do for `Commit()` is batching trie changes per txn, and not per block (thus smaller). Is that fine? (just double checking)

What this means for serialization here, is that when `BatchSerialize(..)` is called, we already lost all the `cows` already so we can't use them for anything as suggested.

[Maybe we can think of a way to keep some sort of "cow" that lives from the trie opening until the serialization or something like that. But I'd like first to get your thoughts before going deeper on potential solutions]

tree.go
jsign: Good point. I'd have to think a bit about how to reorganize some things here, maybe with an extra `map`.
I'll think about it a bit.

jsign:
> 1. I think there is some redundancy since, in order to compute the commitment to the parent, the child's commitment already has to be serialized. Caching the serialization done during the commit and reusing it during the parent's serialization could prove to be quite a gain.

When we commit to the parent, child commitments aren't "serialized", at least in the semantics of "serialize" that happens in `BatchSerialize()`.

When we Commit(), what we do in the children is a Point->Fr (is this what are you referring to serialize cache?).  But I'm not sure caching might be useful here, since what we need to do in `BatchSerialize()` is `ElementToBytes` which isn't related to Point->Fr but Projective->Affine. The former is `X/Y` and the latter is `X/Z` & `Y/Z`.

LMK your thoughts.

> 2. By default, a `HashedNode` doesn't have a deserialized commitment. Only the children that are accessed/resolved are turned into points. But there could be nodes that are deserialized just for reading, and they would have to be serialized again, even though its serialization was already known through the `HashedNode`'s `commitment` field, which is a `[]byte`. This can potentially save a lot of computations when reading code/storage. And line 1343, you would iterate over `n.cow` instead of `n.children`.

Touched on this in a PR comment reply.

tree.go
jsign: Because `bitlist` is used and filled before `ret` is defined. 

To define `ret` here in L1369 we use/need `nhashed` (to calculate the length). And we calculate `nhashed` when constructing `bitlist` (L1361).

What we could do is move `ret` to line ~L1358, and maybe assume a `len(compressedChildren)+nhashed` to be a worst case scenario, so we don't need the exact `nhashed`. After we have `nhashed` we can redefine `ret` to have this exact length again, and do the same.

The only tradeoff is reserving some extra unneeded capacity, but that can avoid the `copy` here (which is a 32 byte copy, so not sure if will make a difference but we can try).

jsign: OK, implemented the idea to be able to do this. 32 bytes isn't that bad to copy, but also some extra bytes in memory shouldn't hurt.

Done!

tree.go
```diff
@@ -1368,20 +1361,26 @@ func (n *InternalNode) serializeWithCompressedChildren(compressedChildren [][32]
...
 	consumed := 0
 	for i := 0; i < NodeWidth; i++ {
 		if bit(bitlist, i) {
 			if bit(hashlist[:], i) {
 				children = append(children, n.children[i].(*HashedNode).commitment...)
 			} else {
-				children = append(children, compressedChildren[consumed][:]...)
+				childIdx, ok := compressedPointsIdxs[n.children[i]]
+				if !ok {
+					panic("children commitment not found in cache")
```
jsign, line 1381: This should never happen, but I'm not a fan of panics (which we already have in many other places).

Maybe it's worth, after things stabilize, to reconsider all panics and always return errors. In any case, the "error" is fatal anyway since go-ethereum won't be able to recover so not sure.

tree.go
jsign: OK, [fixed it here](https://github.com/gballet/go-verkle/pull/333/commits/b1e3111371b05643f4ee49ab53830d69889b7878).
Maybe take a look if looks convincing.

tree.go
jsign: Circling this a bit more looks like our replay benchmark doesn't detect our "past byzantium" hard fork, which I remember you mentioned some months ago this "intermediate root" calculation was removed.

I [double checked here](https://github.com/gballet/go-ethereum/blob/7d1fb48778c3e2b93913738708c66d5e5bf8f2ec/core/state_processor.go#L114), and in the replay run it takes the `else` route.

So this is a bit pessimistic for our intentions since the more we can batch, the better our optimizations would work compared with MPT (which probably doesn't care that much about "batching" as VKT crypto).

gballet (approved):
It has been extensively tested by now, merging.

PR #336, opened 2023-03-20
README: fix precomp file referencing

jsign:
The current readme was still directed to an old release. I changed a bit the sentence to avoid remembering to update this link.
If we want to still keep the concrete link in the README, I can change it (no strong opinion on that).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Node width is set to 256 children.
 
 ### Setup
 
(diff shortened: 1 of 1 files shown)


gballet (approved):
LGTM. Waiting for github.com/gballet/go-ethereum#180 to be merged before merging this one, so that both branches remain in sync.

PR #337, opened 2023-03-21
batch frs in leaf node updating & batch proof items sections

jsign:
I did a fresh pass looking for more `toFr(...)` that can have batching opportunities, and I think these are the ones we were missing.

This had some slight impact in our current replay benchmark, but maybe can be more relevant when we'll have fuller blocks in the next version.

Before:
image

After:
image

We have net savings.


tree.go
```diff
@@ -223,10 +223,9 @@ func NewLeafNode(stem []byte, values [][]byte) *LeafNode {
 
 	count = fillSuffixTreePoly(c1poly[:], values[:128])
 	leaf.c1 = cfg.CommitToPoly(c1poly[:], 256-count)
-	toFr(&poly[2], leaf.c1)
 	count = fillSuffixTreePoly(c2poly[:], values[128:])
 	leaf.c2 = cfg.CommitToPoly(c2poly[:], 256-count)
-	toFr(&poly[3], leaf.c2)
+	toFrMultiple([]*Fr{&poly[2], &poly[3]}, []*Point{leaf.c1, leaf.c2})
```
jsign, line 228: Low hanging fruit here.

tree.go
```diff
@@ -723,10 +722,14 @@ func (n *InternalNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]
...
-	fi := make([]Fr, NodeWidth)
+	var fi [NodeWidth]Fr
+	var fiPtrs [NodeWidth]*Fr
+	var points [NodeWidth]*Point
 	for i, child := range n.children {
-		toFr(&fi[i], child.Commitment())
+		fiPtrs[i] = &fi[i]
+		points[i] = child.Commitment()
 	}
+	toFrMultiple(fiPtrs[:], points[:])
```
jsign, line 732: Proof related: Interesting gains the "fuller" internal node.

tree.go
```diff
@@ -903,29 +906,16 @@ func (n *LeafNode) insertMultiple(k []byte, values [][]byte) error {
...
-		oldc Fr
-	)
-	if index < 128 {
-		c = n.c1
-	} else {
-		c = n.c2
-	}
-	toFr(&oldc, c)
-	return c, &oldc
-}
```
jsign, line 918: It went away since it was forcing us to `toFr` individually.

tree.go
```diff
@@ -903,29 +906,16 @@ func (n *LeafNode) insertMultiple(k []byte, values [][]byte) error {
...
+	// Calculate the Point-delta.
+	var poly [NodeWidth]Fr
+	poly[cxIndex] = deltaC
 
-	toFr(&newc, c)
-	newc.Sub(&newc, oldc)
-	poly[2+(index/128)] = newc
+	// Add delta to the current commitment.
 	n.commitment.Add(n.commitment, cfg.CommitToPoly(poly[:], 0))
 }
```
jsign, line 920: The previous version had L926 which was forcing us to do single Fr transformation.
Changed a bit here to:
- Receive `newC` and `oldC` via params, since the caller will exploit doing some batching on his side.
- Renamed vars and added some comments to be clearer.
- Changed `index` to `cxIndex` semantics, since this method was always called with params `0` and `128` which was a bit confusing. No need to do L928, just let caller tell us directly the index. On the caller side will be clear, see later.

tree.go
```diff
@@ -958,41 +948,77 @@ func (n *LeafNode) updateCn(index byte, value []byte, c *Point) {
...
+	// Batch the Fr transformation of the new and old CX.
+	var frs [2]Fr
+	toFrMultiple([]*Fr{&frs[0], &frs[1]}, []*Point{c, &oldC})
+
+	// If index is in the first NodeWidth/2 elements, we need to update C1. Otherwise, C2.
+	cxIndex := 2 + int(index)/(NodeWidth/2) // [1, stem, -> C1, C2 <-]
+	n.updateC(cxIndex, frs[0], frs[1])
 
 	n.values[index] = value
 }
```
jsign, line 972: Single leaf node updating new batching:
- First capture corresponding `c` and `oldC`.
- Batch Fr calc on both, since before we did two separate `toFr`.
- L968 and surrounding comments are paranoid explicit on "CX index" logic.

tree.go
```diff
@@ -958,41 +948,77 @@ func (n *LeafNode) updateCn(index byte, value []byte, c *Point) {
...
 
 func (n *LeafNode) updateMultipleLeaves(values [][]byte) {
```
jsign, line 974: Some changes in the method where we may update multiple leaf values.
Apart from code changes, I left some comments for future readers to understand _what_ and _why_ we're doing this.

tree.go
```diff
@@ -958,41 +948,77 @@ func (n *LeafNode) updateCn(index byte, value []byte, c *Point) {
...
 
-	if c1 != nil {
-		n.updateC(0, c1, old1)
-	}
-	if c2 != nil {
-		n.updateC(128, c2, old2)
+	// We have three potential cases here:
+	// 1. We have touched C1 and C2: we Fr-batch old1, old2 and newC1, newC2. (4x gain ratio)
+	// 2. We have touched only one CX: we Fr-batch oldX and newCX. (2x gain ratio)
+	// 3. No C1 or C2 was touched, this is a noop.
```
jsign, line 1007: Depending on the case, we try to exploit batching as much as we can.

gballet (approved):
Looks good in general, I'm just wondering if we ever considered using a `[]*Fr` as a parameter to `CommitToPoly` ? All these copies must have a cost. Not as much as doing individual `toFr`s of course.

PR #338, opened 2023-04-06
VKT state migration optimized methods & less aggressive LeafNode memory pressure

jsign:
This PR contains:
- A new `conversion.go` file with specifically crafted methods to support `geth` VKT state migration subcommand.
- A change in value representation in `LeafNodes` optimized for the conversion case where spare 256 slices create a lot of memory pressure for the VKT conversion. This might be beneficial too for "daily" usage, but doubt it might be entirely useful for that case.


go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/gballet/go-verkle
 
 go 1.18
 
-require github.com/crate-crypto/go-ipa v0.0.0-20230315201338-1643fdc2ead8
+require github.com/crate-crypto/go-ipa v0.0.0-20230410135559-ce4a96995014
```
jsign, line 5: Basically applying the same change here https://github.com/gballet/go-verkle/pull/339.

tree.go
```diff
@@ -249,7 +262,7 @@ func (n *InternalNode) Children() []VerkleNode {
 }
 
 func (n *InternalNode) SetChild(i int, c VerkleNode) error {
-	if i >= NodeWidth-1 {
+	if i >= NodeWidth {
```
jsign, line 265: I forgot to mention: I found this bug the hard way while doing the helpers and running stuff converting trees...

I think we got lucky to not see this bug happening before.

conversion.go
```diff
@@ -0,0 +1,196 @@
...
+
+		valsslice := make([][]byte, NodeWidth)
+		for idx := range nv.Values {
+			valsslice[idx] = nv.Values[idx]
+		}
+
+		fillSuffixTreePoly(c1poly[:], valsslice[:NodeWidth/2])
+		ret[i].c1 = cfg.CommitToPoly(c1poly[:], 0)
+		fillSuffixTreePoly(c2poly[:], valsslice[NodeWidth/2:])
+		ret[i].c2 = cfg.CommitToPoly(c2poly[:], 0)
```
jsign, line 40: Now that we merged the other emptyCode cached PR, we can exploit the same here. Maybe I can extract this code section to be shared between both.

For now, it won't make a big difference in the conversion since we're bottleneck by compactions.
I prefer to merge this PR as is with our correctness guarantees, and I'll do a PR right after with only that refactor.

jsign:
> Looks good overall, but isn't using a map slowing things down, especially the serialization part?

This change was optimized for the tree conversion. In that case, we have a lot of memory pressure by having 256-slices where, most of the nodes, only use a few slots. The more efficient the RAM use, the more batching we can do.

For replay benchmark or "usual" run of block execution, that memory waste isn't that big since we don't have that many nodes in memory. In that case, using a map is paying some CPU cost for the cryptographic hash function _maybe_ unnecessarily. I'd say is I change I'd ideally want to keep for the conversion code runs, and maybe not for master.

jsign:
Marking this PR `on-hold` and switching to draft to avoid merging.

This PR is complete and works for a full tree conversion, but now we're also exploring in parallel the overlay tree conversion. 
After we figure out which strategy we'll end up using, we should hold a bit merging this.

PR #340, opened 2023-04-10
Cache EmptyCode hash point in polynomial commitment

jsign:
This PR implements an optimization suggested by Guillaume around empty code hashes.

The rationale is that EOA addresses must store an “empty code” hash; thus, a reasonable amount of “account headers” will have this particular value in the “code hash” slot.

I decided to do this optimization in this repository since go-ipa is a more general cryptographic library. This kind of optimization is closer to exploiting a fact about the domain, not a “cryptography” optimization.

The way this works is quite simple: in the init phase, the empty code hash is precomputed. The overhead is 7 microseconds and 64 bytes of RAM, so it sounds light to be considered part of a precomp file and introduce complexity (which involves a separate repo BTW).

I also included a new benchmark to compare the optimized case vs any other case:

```bash
goos: linux
goarch: amd64
pkg: github.com/gballet/go-verkle
cpu: AMD Ryzen 7 3800XT 8-Core Processor            
BenchmarkEmptyHashCodeCachedPoint/emptyHash-16              2518101               471.5 ns/op            96 B/op          1 allocs/op
BenchmarkEmptyHashCodeCachedPoint/nonEmptyHash-16            160844              7073 ns/op              96 B/op          1 allocs/op
PASS
ok      github.com/gballet/go-verkle    4.185s
```

This is benchmarking a 256-vector with all zero elements but the code hash slots. The speedup is significant but can be made irrelevant by a vector with other nonzero values. Saying it differently, we’re saving 6 microseconds per “account header” computation; that’s a better way to think about it.

I ran our current replay benchmark with this optimization and didn’t see a relevant difference in total running time. So, those CPU savings are there, but in the grand scheme of things, these 7 microseconds per account header didn’t seem to be noticeable (e.g: db compactions have some order of magnitude more difference in wallclock time).

In any case, I think it’s a reasonable optimization to keep since it isn’t too complex and might become more impactful in maybe other scenarios which are more strictly CPU bound than importing a chain or doing a migration.

To be sure about correctness, an extra test compares the result using the current cached point optimization against a result obtained from current `master`.

### DIFF
--- a/config_ipa.go
+++ b/config_ipa.go
@@ -26,9 +26,24 @@
 package verkle
 
 import (
+	"encoding/hex"
 	"os"
 
 	"github.com/crate-crypto/go-ipa/ipa"
+)
+
+// emptyCodeHashPoint is a cached point that is used to represent an empty code hash.
+// This value is initialized once in GetConfig().
+var (
+	emptyCodeHashPoint           *Point
+	emptyCodeHashFirstHalfValue  Fr
+	emptyCodeHashSecondHalfValue Fr
+)
+
+const (
+	codeHashVectorPosition     = 3 // Defined by the spec.
+	emptyCodeHashFirstHalfIdx  = codeHashVectorPosition * 2
+	emptyCodeHashSecondHalfIdx = emptyCodeHashFirstHalfIdx + 1
 )
 
 type IPAConfig struct {
@@ -38,7 +53,23 @@ type IPAConfig struct {
 type Config = IPAConfig
 
 func (ipac *IPAConfig) CommitToPoly(poly []Fr, _ int) *Point {
+	containsEmptyCodeHash := len(poly) >= emptyCodeHashSecondHalfIdx &&
+		poly[emptyCodeHashFirstHalfIdx].Equal(&emptyCodeHashFirstHalfValue) &&
+		poly[emptyCodeHashSecondHalfIdx].Equal(&emptyCodeHashSecondHalfValue)
+
+	if containsEmptyCodeHash {
+		poly[emptyCodeHashFirstHalfIdx] = FrZero
+		poly[emptyCodeHashSecondHalfIdx] = FrZero
+	}
+
 	ret := ipac.conf.Commit(poly)
+
+	if containsEmptyCodeHash {
+		ret.Add(&ret, emptyCodeHashPoint)
+		poly[emptyCodeHashFirstHalfIdx] = emptyCodeHashFirstHalfValue
+		poly[emptyCodeHashSecondHalfIdx] = emptyCodeHashSecondHalfValue
+	}
+
 	return &ret
 }
 
@@ -54,7 +85,7 @@ func GetConfig() *Config {
 			serialized, err := ipacfg.SRSPrecompPoints.SerializeSRSPrecomp()
 			if err != nil {
 				panic("error writing serialized precomputed Lagrange points:" + err.Error())
-			} else if err = os.WriteFile(precompFileName, serialized, 0666); err != nil {
+			} else if err = os.WriteFile(precompFileName, serialized, 0o666); err != nil {
 				panic("error saving the precomp: " + err.Error())
 			}
 		} else {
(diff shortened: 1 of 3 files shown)


config_ipa.go
```diff
@@ -26,19 +26,50 @@
...
 type IPAConfig struct {
 	conf *ipa.IPAConfig
 }
 
 type Config = IPAConfig
 
 func (ipac *IPAConfig) CommitToPoly(poly []Fr, _ int) *Point {
+	containsEmptyCodeHash := len(poly) >= emptyCodeHashSecondHalfIdx &&
+		poly[emptyCodeHashFirstHalfIdx].Equal(&emptyCodeHashFirstHalfValue) &&
+		poly[emptyCodeHashSecondHalfIdx].Equal(&emptyCodeHashSecondHalfValue)
```
jsign, line 58: We detect if the empty code hash is present in the right slots of the vector.

config_ipa.go
```diff
@@ -26,19 +26,50 @@
...
+		poly[emptyCodeHashSecondHalfIdx] = FrZero
+	}
+
 	ret := ipac.conf.Commit(poly)
+
+	if containsEmptyCodeHash {
+		ret.Add(&ret, emptyCodeHashPoint)
+		poly[emptyCodeHashFirstHalfIdx] = emptyCodeHashFirstHalfValue
+		poly[emptyCodeHashSecondHalfIdx] = emptyCodeHashSecondHalfValue
+	}
```
jsign, line 71: If that's the case (i.e: `containsEmptyCodeHash == true`), then we zero-out the appropriate slots, calculate the commitment (so those slots will be ignored thus avoiding computation), and after we get the result we add the cached point (L68). Note that we also restore the original values in L69 and L70; this is important to keep `CommitToPoly` not mutate what the caller sent in the `poly` parameter.

config_ipa.go
```diff
@@ -65,12 +96,24 @@ func GetConfig() *Config {
...
 		cfg = &IPAConfig{conf: ipacfg}
+
+		emptyHashCode, _ := hex.DecodeString("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470")
+		values := make([][]byte, NodeWidth)
+		values[codeHashVectorPosition] = emptyHashCode[:]
+		var c1poly [NodeWidth]Fr
+		fillSuffixTreePoly(c1poly[:], values[:NodeWidth/2])
+		emptyCodeHashPoint = cfg.CommitToPoly(c1poly[:], 0)
+		emptyCodeHashFirstHalfValue = c1poly[emptyCodeHashFirstHalfIdx]
+		emptyCodeHashSecondHalfValue = c1poly[emptyCodeHashSecondHalfIdx]
```
jsign, line 107: As part of the precomp setup, we calculate the cached point. We leverage the same methods in the VKT data structure, so things should be coherent.

tree_test.go
```diff
@@ -1168,3 +1168,62 @@ func TestRustBanderwagonBlock48(t *testing.T) {
...
+func BenchmarkEmptyHashCodeCachedPoint(b *testing.B) {
+	_ = GetConfig()
+	const codeHashVectorPosition = 3 // Defined by the spec.
+
+	testCases := []struct {
+		name     string
+		hashCode string
+	}{
+		{name: "emptyHash", hashCode: "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"},
+		{name: "nonEmptyHash", hashCode: "4242420186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"},
```
jsign, line 1181: We compare the empty code hash against a slightly modified hash that wouldn't leverage the cached point, which allows seeing in the output the performance difference between the optimized and unoptimized case.

gballet (dismissed):
LGTM. I'm a bit surprised that the improvement is not more impressive than that, especially for the first step of the conversion.

gballet:
@jsign before I merge, could you please fix the deepsource error?

gballet:
Come to think of it, I believe that there is a simpler way of doing that: when the empty code hash is created, it's always during the creation, which means that it could be captured in `NewLeafNode`, e.g. here: https://github.com/gballet/go-verkle/pull/338/files#diff-050830ac4334170e8d335e1b95519052a3461c5b57a4ca38a0531e35a8f3d388R207

If it is the case, then `c1` would be initialized with the empty code point instead of `Generator`, and it would make the total footprint of this PR smaller, while also not checking two values of the polynomial each time. What do you think?

jsign:
> 

I like the idea! Since it avoids the two-byte checking for polynomial commitments related to internal nodes, which will ~always be false.

I'll adjust and mark it to review again

tree.go
```diff
@@ -204,32 +204,49 @@ func New() VerkleNode {
 
 // New creates a new leaf node
 func NewLeafNode(stem []byte, values [][]byte) *LeafNode {
```
jsign, line 206: Some notes here:
- I moved the "value detection" to this method now. 
- I refactored the overall previous code to make it clearer (e.g: the `Generator()` previous calls weren't necessary (not even for `c2`), narrow the scope of some variables, etc).

jsign (commented):
@gballet, feel free to give it another look.
(I resolved all previous comments to only leave the new one)

PR #342, opened 2023-04-14
tree: fix global pointer mutation

jsign:
This PR fixes a bug in an unwanted pointer copy and mutation of a global var which should be read only.

### DIFF
--- a/config_ipa.go
+++ b/config_ipa.go
@@ -35,7 +35,7 @@ import (
 // EmptyCodeHashPoint is a cached point that is used to represent an empty code hash.
 // This value is initialized once in GetConfig().
 var (
-	EmptyCodeHashPoint           *Point
+	EmptyCodeHashPoint           Point
 	EmptyCodeHashFirstHalfValue  Fr
 	EmptyCodeHashSecondHalfValue Fr
 )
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -208,14 +208,13 @@ func NewLeafNode(stem []byte, values [][]byte) *LeafNode {
 		c1poly[EmptyCodeHashFirstHalfIdx].Equal(&EmptyCodeHashFirstHalfValue) &&
 		c1poly[EmptyCodeHashSecondHalfIdx].Equal(&EmptyCodeHashSecondHalfValue)
 	if containsEmptyCodeHash {
-		// We start c1 with the empty code hash point.
-		c1 = EmptyCodeHashPoint
```
jsign, line 212: Before `EmptyCodeHashPoint` was a `*Point`, so this was doing a pointer copy and then in L218 (deleted line) it was mutating the value of that global var. 

Pretty shocking this didn't become apparent in tests in this repo nor a different hash in the replay bench. :/

config_ipa.go
```diff
@@ -35,7 +35,7 @@ import (
 // EmptyCodeHashPoint is a cached point that is used to represent an empty code hash.
 // This value is initialized once in GetConfig().
 var (
-	EmptyCodeHashPoint           *Point
+	EmptyCodeHashPoint           Point
```
jsign, line 38: See the below comment to understand more about this.
But I changed this to be sure this kind of bug won't happen again, and we copy global values and not pointers which is safer.

tree.go
gballet: indeed

gballet (approved):
LGTM

PR #343, opened 2023-04-19
Overlay tree migration explorations

jsign:
TL;DR: This PR contains a test/benchmark that simulates a touched tree with X key/values, and a later stage of migration Y key/values from a base tree. It includes some optimized methods for the migrated key/values. 

The main intention of this PR is to show the exploration of:
- Using the normal `Insert(.., ..)` API for the migrated key/values, which is somewhat naive compared to an optimized strategy, comparing speedups.
- Already play with the optimized strategy and the "merging logic".
- Simply write about findings/thoughts/next steps.

This PR isn't meant to be merged necessarily, so I'll keep it as a draft.
Not expecting a formal review, if you want to skim the code I'd recommend looking at:
- `tree_test.go` to see the new test/benchmark; you'll get 90% of the gist of the story there.
- `conversion.go` see the new method `InsertMigratedLeaves`.

This code isn't final or similar, so expect to see rough edges.

---

I created a test/benchmark that compares some scenarios between using an “unbatched” insertion for the migrated key/vals (i.e: `Insert(key, value)` in a nutshell), and a “batched” version which uses some work I did for the full-tree migration version with some twists and extra stuff (nothing wild, just reasonable ideas).

Sharing the output of this new test/benchmark. ’ll add some notes below on how to read this: 

```
=== RUN   TestBatchInsertOrdered
Assuming 0 key/values touched by block execution:
        If 1000 extra key-values are migrated: unbatched 44ms, batched 24ms, 1.83x
        If 2000 extra key-values are migrated: unbatched 79ms, batched 41ms, 1.93x
        If 5000 extra key-values are migrated: unbatched 200ms, batched 93ms, 2.14x
        If 8000 extra key-values are migrated: unbatched 317ms, batched 147ms, 2.16x
Assuming 500 key/values touched by block execution:
        If 1000 extra key-values are migrated: unbatched 49ms, batched 30ms, 1.61x
        If 2000 extra key-values are migrated: unbatched 86ms, batched 47ms, 1.80x
        If 5000 extra key-values are migrated: unbatched 205ms, batched 100ms, 2.03x
        If 8000 extra key-values are migrated: unbatched 320ms, batched 153ms, 2.09x
Assuming 1000 key/values touched by block execution:
        If 1000 extra key-values are migrated: unbatched 54ms, batched 37ms, 1.46x
        If 2000 extra key-values are migrated: unbatched 93ms, batched 53ms, 1.73x
        If 5000 extra key-values are migrated: unbatched 212ms, batched 107ms, 1.98x
        If 8000 extra key-values are migrated: unbatched 330ms, batched 161ms, 2.04x
Assuming 2000 key/values touched by block execution:
        If 1000 extra key-values are migrated: unbatched 69ms, batched 50ms, 1.37x
        If 2000 extra key-values are migrated: unbatched 106ms, batched 67ms, 1.59x
        If 5000 extra key-values are migrated: unbatched 226ms, batched 122ms, 1.85x
        If 8000 extra key-values are migrated: unbatched 348ms, batched 174ms, 2.00x
Assuming 5000 key/values touched by block execution:
        If 1000 extra key-values are migrated: unbatched 106ms, batched 90ms, 1.17x
        If 2000 extra key-values are migrated: unbatched 146ms, batched 108ms, 1.36x
        If 5000 extra key-values are migrated: unbatched 265ms, batched 163ms, 1.63x
        If 8000 extra key-values are migrated: unbatched 407ms, batched 235ms, 1.73x
```

So, look at:

```bash
Assuming X key/values touched by block execution:
        If Y extra key-values are migrated: unbatched Ams, batched Bms, Cx
```

 Then X, Y, A, B, and C mean:

- X is the number of “active/touched” key/values in the tree in memory. X is simulating some touched key/values after a block execution.
- Y is the number of key/values that we migrate from the MPT. They’re randomly generated.
- A is the number of milliseconds it takes to insert the migrated key/values in the scenario for the “naive” strategy.
- B is the number of milliseconds it takes to insert the migrated key/values in the scenario for the optimized strategy.
- C is the speedup comparing the naive vs optimized case for the same scenario.

To understand what these “milliseconds it takes to insert the migrated/key values in this scenario”, the high-level run of the scenario is:

1. Generate a VKT with X random key/values
2. Start counting time.
3. Insert the Y key/values
4. Commit the tree.
5. Serialize the tree
6. Stop counting time.
[This is run both with naive (unbatched) API, and the optimized (batched) API; so we get A and B times, respectively]

The time between 2 and 6 is what I report as that time. It’s a bit “unfair” in that committing and serializing the tree also involves doing the work for the X key/values, which isn’t only work related to the Y key/values. But we can’t separate them since commit and serialize exploits batching in everything that exists (which is actually a good thing). So the actual speedup of the optimized implementation is probably better. In any case, it is good to have a number.

What we can conclude from this:
- The optimized version makes inserting the migrated key/values cheaper in relative terms with “usual” insertions from a block execution since we can exploit some things to insert them more efficiently.
- The “millisecond times” (B value), can give a rough sense of pure CPU work for inserting+committing+serializing which is most/all of the CPU work.
- Reasonable confidence that the optimized version is correct since in all scenarios I’m comparing that the resulting root of both VKT with the unoptimized and optimized is the same. (That's why I mention this is a test/benchmark; doing both)

What we can’t conclude from this:
- As mentioned, this is mostly measuring CPU work.
- In a “full scenario”, inserting key/values of migrated MPT data will involve resolving some `HashedNode` in the path to insertion, which implies some extra disk lookups to load those nodes. Nothing weird, and quite normal, but have it in mind. Probably for the first 2-3 layers we don’t need new disk accesses, so it shouldn’t be that bad.
- In a “full scenario” the base VKT tree would be probably a bit deeper; but that’s handwavy since it depends on how far into the full MPT migration you’re in, actually.

What we could do about the “can’t conclude” points is actually “embedding” this same scenario test in our replay benchmark with real data, which would be very interesting since:
- The “touched” tree will be real, since it’s a real block execution. Or “as real” or “updated” as the data we use for importing the chain.
- We can still generate random key/values to simulate MPT data; since that would be a worst-case anyway.
- We can fully resolve `HashedNodes` for real, involving the mentioned disk lookups to load parts of the tree that aren’t in memory.
- We can compare the “slowdown” or “extra work” for the replay benchmark with and without migration, tunning for different Y values (number of keys migrated per block).


conversion.go
```diff
@@ -0,0 +1,152 @@
...
+
+// BatchNewLeafNodeData is a struct that contains the data needed to create a new leaf node.
+type BatchNewLeafNodeData struct {
+	Stem   []byte
+	Values map[byte][]byte
+}
+
+// BatchNewLeafNode creates a new leaf node from the given data. It optimizes LeafNode creation
+// by batching expensive cryptography operations. It returns the LeafNodes sorted by stem.
+func BatchNewLeafNode(nodesValues []BatchNewLeafNodeData) []LeafNode {
```
jsign, line 19: I'm using here something I did for the full tree conversion, but added parallelization at this layer. This wasn't needed in the tree conversion since, in that case we parallelized "at the client level" by working in subtrees.

But the idea is the same/similar.

conversion.go
```diff
@@ -0,0 +1,152 @@
...
+func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
+	for i := range stem1 {
+		if stem1[i] != stem2[i] {
+			return i
+		}
+	}
+	panic("stems are equal")
+}
+
+func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
```
jsign, line 85: This is a sketch/first-version of a method that receives prepared `LeafNode` of base tree key/values, and attemps to "merge them" in a living VKT. Can have rough edges, just consider it skimmable.

conversion.go
```diff
@@ -0,0 +1,152 @@
...
+				}
+				n.children[ln.stem[parent.depth]] = resolved
+			}
+
+			nextParent, ok := parent.children[ln.stem[parent.depth]].(*InternalNode)
+			if !ok {
+				break
+			}
+			parent = nextParent
+		}
```
jsign, line 109: TL;DR: Walk the tree to get the place where the `LeafNode` parent should be. The stopping point is not finding an `InternalNode` (if `HashedNode`, resolve and keep walking).

conversion.go
```diff
@@ -0,0 +1,152 @@
...
+
+		switch node := parent.children[ln.stem[parent.depth]].(type) {
+		case Empty:
+			parent.cowChild(ln.stem[parent.depth])
+			parent.children[ln.stem[parent.depth]] = &ln
+			ln.setDepth(parent.depth + 1)
```
jsign, line 115: Easy case, insert leaf and call it a day.

conversion.go
```diff
@@ -0,0 +1,150 @@
...
+		case *LeafNode:
```
jsign, line 116: OK, this case is more interesting. We have two subcases.

conversion.go
```diff
@@ -0,0 +1,150 @@
...
+				parent.children[ln.stem[parent.depth]] = nextParent
+				parent = nextParent
+			}
+			// Add old and new leaf node to the latest created parent.
+			parent.cowChild(node.stem[parent.depth])
+			parent.children[node.stem[parent.depth]] = node
+			ln.setDepth(parent.depth + 1)
+			parent.cowChild(ln.stem[parent.depth])
+			parent.children[ln.stem[parent.depth]] = &ln
+			ln.setDepth(parent.depth + 1)
```
jsign, line 146: This is the most probable case. There's a Leaf with a different stem.
We do what we should: create a path of internal nodes to "split" the tree accordingly, and insert our migrated&prepared Leaf.

conversion.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	panic("stems are equal")
+}
+
+func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
+	for i := range leaves {
+		ln := leaves[i]
+		parent := n
+
+		// Look for the appropiate parent for the leaf node.
```
jsign, line 90: Wal the tree looking for the parent for the LeafNode to be inserted. While we walk, we have to resolve potential `HashedNodes`. Note that we carefully (L109) mark our walk as cow-ed since we know we'll insert the leaf "down the road".

conversion.go
```diff
@@ -0,0 +1,159 @@
...
+				if err != nil {
+					return fmt.Errorf("parsing node %x: %w", serialized, err)
+				}
+				parent.children[ln.stem[parent.depth]] = resolved
+			}
+
+			nextParent, ok := parent.children[ln.stem[parent.depth]].(*InternalNode)
+			if !ok {
+				break
+			}
```
jsign, line 107: If we find a LeafNode or Empty, we're done; we have the parent.

conversion.go
```diff
@@ -0,0 +1,159 @@
...
+			// Add old and new leaf node to the latest created parent.
+			parent.cowChild(node.stem[parent.depth])
+			parent.children[node.stem[parent.depth]] = node
+			node.setDepth(parent.depth + 1)
+			parent.cowChild(ln.stem[parent.depth])
+			parent.children[ln.stem[parent.depth]] = &ln
+			ln.setDepth(parent.depth + 1)
+		default:
+			return fmt.Errorf("unexpected node type %T", node)
```
jsign, line 155: Just be sure nothing else can be found, which would be a bug.

conversion.go
```diff
@@ -0,0 +1,159 @@
...
+				node.updateMultipleLeaves(nonPresentValues)
+				continue
+			}
+
+			// Otherwise, we need to create the missing internal nodes depending in the fork point in their stems.
+			idx := firstDiffByteIdx(node.stem, ln.stem)
+			// We do a sanity check to make sure that the fork point is not before the current depth.
+			if byte(idx) <= parent.depth {
+				return fmt.Errorf("unexpected fork point %d for nodes %x and %x", idx, node.stem, ln.stem)
+			}
```
jsign, line 139: Let's be paranoid and check if this invariant holds; if doesn't hold, it's a bug.

conversion.go
```diff
@@ -0,0 +1,159 @@
...
+				parent.children[ln.stem[parent.depth]] = nextParent
+				parent = nextParent
+			}
+			// Add old and new leaf node to the latest created parent.
+			parent.cowChild(node.stem[parent.depth])
+			parent.children[node.stem[parent.depth]] = node
+			node.setDepth(parent.depth + 1)
+			parent.cowChild(ln.stem[parent.depth])
+			parent.children[ln.stem[parent.depth]] = &ln
+			ln.setDepth(parent.depth + 1)
```
jsign, line 153: Create the needed internal points depending on the "fork section" of the steam, and connect the final parent with the existing LeafNode and the "to be inserted" LeafNode.

conversion.go
```diff
@@ -0,0 +1,159 @@
...
+	panic("stems are equal")
+}
+
+func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
+	for i := range leaves {
+		ln := leaves[i]
+		parent := n
+
+		// Look for the appropiate parent for the leaf node.
+		for {
```
gballet, line 91: I'd prefer a recursive version of this in the end, but as long as it works fine there's no issue for now.

conversion.go
```diff
@@ -0,0 +1,150 @@
...
+				nonPresentValues := make([][]byte, NodeWidth)
+				for i := range ln.values {
+					if node.values[i] == nil {
+						nonPresentValues[i] = ln.values[i]
+					}
+				}
+
+				node.updateMultipleLeaves(nonPresentValues)
+				continue
+			}
```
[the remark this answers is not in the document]
jsign: Yep.

tree.go
```diff
@@ -259,7 +259,7 @@ func (n *InternalNode) Children() []VerkleNode {
 }
 
 func (n *InternalNode) SetChild(i int, c VerkleNode) error {
-	if i >= NodeWidth-1 {
+	if i >= NodeWidth {
```
jsign, line 262: This was a bug.

gballet (approved):
LGTM

PR #344, opened 2023-04-19
add a helper to dump a tree to JSON

gballet:
This is a request from Nebula, working on spec tests. This adds a helper function that you can call on any internal node to serialize it to JSON.


jsign (approved):
LGTM

PR #345, opened 2023-04-19
use nil in InternalNode to turn it into a stateless node

gballet:
This is a first attempt at getting rid of `StatlessNode` in order to simplify the code and be able to use all the optimizations that were made recently, in a statless context.

Currently the idea is to use `nil` to signal that a subtree of an `InternalNode` is missing from the stateless view.

I think this should be replaced with `type StatelessSubtree struct{}` or something like that, for safety. I'll sleep on it.


jsign (approved):
LGTM, I mainly trust you until diving into this in the future. :)

PR #346, opened 2023-04-20
safety improvment: use Unknown instead of nil

gballet:


tree.go
```diff
@@ -757,7 +753,9 @@ func (n *InternalNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]
 
 		// Special case of a proof of absence: no children
 		// commitment, as the value is 0.
-		if _, ok := n.children[childIdx].(Empty); ok || n.children[childIdx] == nil {
+		_, isempty := n.children[childIdx].(Empty)
+		_, isunknown := n.children[childIdx].(Unknown)
+		if isempty || isunknown /* this is probably wrong, unknown != empty */ {
```
gballet, line 758: That feels incorrect, but at the same time if the proof goes through an unknown child, it should fail later down the road.

PR #350, opened 2023-05-01
Improve test coverage

jsign:
This PR's main intention is to improve our test coverage. 

As usually happens, looking at coverage profiles, it forces you to see some code in general, and that triggered some overall code improvements like removing duplicated-ish code, adding some extra validation checks, removing unused methods, etc. I’ll explain them in the comments.

Our pre-PR coverage numbers were pretty good. However, 1% of uncovered code in some files could be bad, `tree.go` in particular; this is natural due to organic code evolution. I ensured all important branches are covered now and left some in-comment TODOs in some places I’d like to dive into later to cover appropriately (proof generation/verification related).

Here’re the pre and post-coverage numbers (i.e: `pre% -> post%`):

- Overall coverage: 84.8% → 91.6%
- Per file:
    - config.go (100.0% → 100%)
    - config_ipa.go (74.1% → 88.9%)
    - conversion.go (80.5% → 80.5%)
    - empty.go (61.5% → 92.3%)
    - encoding.go (93.8% → 95.8%)
    - hashednode.go (62.5% → 87.5%)
    - ipa.go (73.3% → 92.3%)
    - proof_ipa.go (91.9% → 91.9%)
    - proof_json.go (80.7% → 80.7%)
    - **tree.go (87.1% → 94.3%)**
    - unknown.go (23.1% → 92.3%)


config_ipa_test.go
```diff
@@ -0,0 +1,31 @@
+package verkle
+
+import (
+	"os"
+	"testing"
+)
+
+func TestGeneratePrecompFile(t *testing.T) {
```
jsign, line 8: This test is double-checking that if the `precomp` file doesn't exist, it's generated with non-zero size.

The need for switching from `const` to `var` in the above file was necessary to be able to test the file generation case.
But things don't stop there, since `cfg` is also global; so there's some extra maneuvering to be sure other tests didn't already populated that variable.

Despite I agree `cfg` being global is somewhat pragmatic since clients don't need to hold `cfg` themselves, it makes some things a bit uglier on this side. From a pure personal taste perspective, I don't like globals and I'm usually fine asking clients to hold their own `cfg`.

In any case, there's no strict need to change this now. Maybe something to make all this safer is to have `GetConfg()` use a `sync.Once` since `GetConfig()` isn't thread-safe. While it's true that might be "part of the contract" that clients should call `GetConfig()` in `main` or something, you never know if some other dependency exists that also uses this package.

This is a bit of an offtopic PR discussion. Not saying we should do something now (or even later), it's not a big deal. Just sharing thoughts.

conversion.go
```diff
@@ -53,7 +53,7 @@ func BatchNewLeafNode(nodesValues []BatchNewLeafNodeData) []LeafNode {
 			var poly [NodeWidth]Fr
 			poly[0].SetUint64(1)
 			for i, nv := range nodesValues {
-				StemFromBytes(&poly[1], nv.Stem)
+				FromLEBytes(&poly[1], nv.Stem)
```
jsign, line 56: I discovered that `StemFromBytes` did the same job as the existing `FromLEBytes`. I think that was a bit confusing since the name seems to imply it does something _special_ related to stems, but in reality, the method doesn't. It is just transforming LE-bytes into a Fr. 

I'll make a PR comment on the exact place where `StemFromBytes` was deleted so we can inspect it a bit better. I'm just mentioning this since you might see other `StemFromBytes -> FromLEBytes` until reaching that point.

debug_test.go
```diff
@@ -33,6 +33,7 @@ func TestJSON(t *testing.T) {
 	root := New()
 	root.Insert(zeroKeyTest, fourtyKeyTest, nil)
 	root.Insert(oneKeyTest, zeroKeyTest, nil)
+	root.Insert(forkOneKeyTest, zeroKeyTest, nil) // Force an internal node in the first layer.
```
jsign, line 36: Tunning a bit more this test to cover a recursion branch.

encoding_test.go
```diff
@@ -22,3 +22,23 @@ func TestLeafStemLength(t *testing.T) {
 		t.Fatalf("invalid serialization when the stem is longer than 31 bytes: %x (%d bytes)", ser, len(ser))
 	}
 }
+
+func TestInvalidNodeEncoding(t *testing.T) {
```
jsign, line 26: Covering some angles in `encoding.go`.

ipa.go
```diff
@@ -69,16 +72,6 @@ func FromBytes(fr *Fr, data []byte) {
 	fr.SetBytes(aligned[:])
 }
 
-func StemFromBytes(fr *Fr, data []byte) {
-	bytes := make([]byte, len(data))
-	copy(bytes, data)
-	fr.SetBytesLE(bytes)
-}
```
jsign, line 76: This method is doing the same as `FromLEBytes`, thus I don't think it's worth having it.
Pushbacks accepted!

ipa.go
```diff
@@ -58,8 +58,11 @@ func toFrMultiple(res []*Fr, ps []*Point) {
...
 
 func FromLEBytes(fr *Fr, data []byte) {
+	if len(data) > 32 {
+		panic("data is too long")
+	}
 	var aligned [32]byte
-	copy(aligned[:len(data)], data)
+	copy(aligned[:], data)
 	fr.SetBytesLE(aligned[:])
 }
```
jsign, line 67: - I added the extra length check since that should hold true. If this doesn't hold, we have a bug somewhere in our logic. Truncating `data` in the `copy` doesn't sound like the right to do in any case.
- I removed that subslice truncation for two reasons: 1. Copy will only copy up to a minimum length of arguments anyway, 2. The `data` length check assumption must never result in `copy` copying less data than expected.

ipa.go
```diff
@@ -69,16 +72,6 @@ func FromBytes(fr *Fr, data []byte) {
...
-	fr.SetBytesLE(bytes)
-}
-
 func Equal(self *Point, other *Point) bool {
 	return other.Equal(self)
 }
-
-func Generator() *Point {
-	return new(Point).Identity()
-}
```
jsign, line 84: This isn't necessary since we merged [this PR](https://github.com/gballet/go-verkle/pull/340/files#diff-050830ac4334170e8d335e1b95519052a3461c5b57a4ca38a0531e35a8f3d388L212-L213).

ipa_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+)
+
+func TestFromBytes(t *testing.T) {
+	var fr Fr
+
+	var beFortyTwo [8]byte
+	binary.BigEndian.PutUint64(beFortyTwo[:], 42)
+
+	FromBytes(&fr, beFortyTwo[:])
```
jsign, line 15: This `FromBytes` method isn't used in `go-verkle` anywhere. It's used in `go-ethereum` for tree-key calculations.
Adding a test to test that it does what we expect.

tree.go
```diff
@@ -301,10 +301,13 @@ func NewLeafNodeWithNoComms(stem []byte, values [][]byte) *LeafNode {
...
+// Children return the children of the node. The returned slice is
+// internal to the tree, so callers *must* consider it readonly.
 func (n *InternalNode) Children() []VerkleNode {
 	return n.children
 }
 
+// SetChild *replaces* the child at the given index with the given node.
 func (n *InternalNode) SetChild(i int, c VerkleNode) error {
 	if i >= NodeWidth {
 		return errors.New("child index higher than node width")
```
jsign, line 313: Both `Children()` and `SetChild()` aren't methods used in `go-verkle`. Tracking down where they were used, they're necessary for the iterator-API in `go-ethereum`.

Added some comments to warn callers that should be mindful about what they're doing with returned parameters, since external parties touching these things can break some invariants in this package.

tree.go
```diff
@@ -313,14 +316,6 @@ func (n *InternalNode) SetChild(i int, c VerkleNode) error {
 	return nil
 }
 
-// TouchCoW is a helper function that will mark a child as
-// "inserted into". It is used by the conversion code to
-// mark reconstructed subtrees as 'written to', so that its
-// root commitment can be computed.
-func (n *InternalNode) TouchCoW(index byte) {
-	n.cowChild(index)
-}
```
jsign, line 322: This wasn't used, so taking the liberty to remove it. I suspect was used in some early approaches of the full tree conversion. I'd prefer to remove this since feels too "hacky" to be a public API.

tree.go
```diff
@@ -393,8 +388,8 @@ func (n *InternalNode) InsertStem(stem []byte, values [][]byte, resolver NodeRes
 		newBranch.children[nextWordInInsertedKey] = leaf
 	case *InternalNode:
 		return child.InsertStem(stem, values, resolver)
-	default: // StatelessNode
-		return errStatelessAndStatefulMix
+	default:
+		return errUnknownNodeType
```
jsign, line 392: @gballet, I think this should be "more correct" now?

tree.go
```diff
@@ -468,6 +463,9 @@ func (n *InternalNode) CreatePath(path []byte, stemInfo stemInfo, comms []*Point
 	return child.CreatePath(path[1:], stemInfo, comms, values)
 }
 
+// GetStem returns the all NodeWidth values of the stem.
+// The returned slice is internal to the tree, so it *must* be considered readonly
+// for callers.
 func (n *InternalNode) GetStem(stem []byte, resolver NodeResolverFn) ([][]byte, error) {
```
jsign, line 469: Interestingly, this `GetStem(...)` was 100% uncovered.

Then I realized that `Get(key)` was having a 99% similar implementation than this method, but that didn't make much sense. `Get(..,...)` a more concrete action than `GetStem(...)`. To avoid duplication, I made `Get(...)` call `GetStem(...)` as you'll see later, and avoid repeated code.

See the next comment.

tree.go
```diff
@@ -585,38 +583,21 @@ func (n *InternalNode) FlushAtDepth(depth uint8, flush NodeFlushFn) {
 	}
 }
 
-func (n *InternalNode) Get(k []byte, getter NodeResolverFn) ([]byte, error) {
-	nChild := offset2key(k, n.depth)
+func (n *InternalNode) Get(key []byte, resolver NodeResolverFn) ([]byte, error) {
```
jsign, line 586: Note how `Get(...)` got simplified. We call `GetStem(...)` and return the value at the needed index.
All the special error-values in different branches are respected by `GetStem(...)` now.

tree.go
```diff
@@ -763,6 +744,7 @@ func (n *InternalNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]
 		if child != nil {
 			points[i] = child.Commitment()
 		} else {
```
jsign, line 746: The next three `// TODO` comments are three branches that are still uncovered.

These are the three branches of `tree.go` that, IMHO, would be good to cover to have "complete basic" coverage of important things. I didn't handle this in this PR since I'm planning to understand more in detail theory and implementation of proof generation/verification in the short-term; so leaving covering these angles at that point.

tree.go
```diff
@@ -912,23 +896,6 @@ func (n *InternalNode) setDepth(d byte) {
...
-			if _, ok := subroot.children[i].(Empty); ok {
-				continue
-			}
-			root.TouchCoW(byte(i))
-			root.children[i] = subroot.children[i]
-		}
-	}
-
-	return root
-}
```
jsign, line 930: I suspect this was used for some early approach of a full-tree conversion strategy.
Looks like a useful method, but if it doesn't seem to be useful today I think it might be better to remove it to avoid adding maintenance costs on code changes. We can always go back in git history and avoid redoing work if needed again.

tree.go
```diff
@@ -937,15 +904,21 @@ func (n *LeafNode) ToHashedNode() *HashedNode {
...
 }
 
-func (n *LeafNode) Insert(k []byte, value []byte, _ NodeResolverFn) error {
+func (n *LeafNode) Insert(key []byte, value []byte, _ NodeResolverFn) error {
+	if len(key) != StemSize+1 {
+		return fmt.Errorf("invalid key size: %d", len(key))
+	}
+	if !bytes.Equal(key[:StemSize], n.stem) {
+		return fmt.Errorf("stems doesn't match: %x != %x", key[:StemSize], n.stem)
+	}
```
jsign, line 913: I am a bit tighter on assumptions here.

tree_test.go
```diff
@@ -44,10 +44,11 @@ import (
 var testValue = []byte("0123456789abcdef0123456789abcdef")
 
 var (
-	zeroKeyTest, _   = hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000000")
-	oneKeyTest, _    = hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000001")
-	fourtyKeyTest, _ = hex.DecodeString("4000000000000000000000000000000000000000000000000000000000000000")
-	ffx32KeyTest, _  = hex.DecodeString("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
+	zeroKeyTest, _    = hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000000")
+	oneKeyTest, _     = hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000001")
+	forkOneKeyTest, _ = hex.DecodeString("0001000000000000000000000000000000000000000000000000000000000001")
```
jsign, line 49: Only adding `forkOneKeyTest` to allow some test cases to force creating internal nodes at depth 1.

tree_test.go
```diff
@@ -285,37 +286,46 @@ func TestDeletePrune(t *testing.T) {
 	key2, _ := hex.DecodeString("0107000000000000000000000000000000000000000000000000000000000000")
```
jsign, line 286: This is a very useful test, I think I never looked at it before.
The essence is the same, but has two extra nits:
- Adds some extra inserted key (`key5`) that puts a value in the same leaf as (`key4`). This is to also cover C2 updating and not only C1 updating.
- Added some extra comments to simply help readers understand what we're doing here. It might be surprising why the hash _shouldn't_ match, which is actually the right thing to happen (because of the "leaf marker").

tree_test.go
```diff
@@ -697,6 +707,10 @@ func TestGetResolveFromHash(t *testing.T) {
 	if !bytes.Equal(data, zeroKeyTest) {
 		t.Fatalf("invalid result: %x != %x", zeroKeyTest, len(data))
 	}
+
+	if hsh := root.Hash(); hsh == nil {
+		t.Fatalf("root hash can't be nil")
+	}
```
jsign, line 713: Coverage.

tree_test.go
```diff
@@ -917,6 +931,39 @@ func TestInsertStem(t *testing.T) {
 	}
 }
 
+func TestInsertStemTouchingBothHalves(t *testing.T) {
```
jsign, line 934: We never had a test for `InsertStem(...)` that would merge new values at both sides of the leaf node values vector, so [we had some case uncovered](https://github.com/gballet/go-verkle/blob/master/tree.go#L1060-L1063).

tree_test.go
```diff
@@ -1278,3 +1325,90 @@ func BenchmarkBatchLeavesInsert(b *testing.B) {
 		}
 	}
 }
+
+func TestManipulateChildren(t *testing.T) {
```
jsign, line 1329: As mentioned, the APIs `Children()` and `SetChild()` aren't used in `go-verkle` anywhere but in iterators in `go-ethereum`. Adding some test cases to cover them.

tree_test.go
```diff
@@ -1278,3 +1325,90 @@ func BenchmarkBatchLeavesInsert(b *testing.B) {
...
+		t.Fatalf("failed to set child")
+	}
+
+	// Verify SetChild() error case.
+	if err := root.(*InternalNode).SetChild(NodeWidth, Empty{}); err == nil {
+		t.Fatalf("expected error")
+	}
+}
+
+func TestLeafNodeInsert(t *testing.T) {
```
jsign, line 1356: Some LeafNode methods were uncovered (i.e: `Value(...)`, `Values(...)` again used in go-ethereum but not generally in go-verkle). Making test tighter to cover them here.

unknown.go
```diff
@@ -52,7 +52,7 @@ func (UnknownNode) Commitment() *Point {
 }
 
 func (UnknownNode) GetProofItems(keylist) (*ProofElements, []byte, [][]byte, error) {
-	panic("trying to produce a commitment for a subtree missing from the stateless view")
+	return nil, nil, nil, errors.New("can't generate proof items for unknown node")
```
jsign, line 55: This felt a bit aggressive, so I'm taking the same policy as we do in `HashedNode`.
LMK if we want to revert!

tree.go
```diff
@@ -937,15 +904,21 @@ func (n *LeafNode) ToHashedNode() *HashedNode {
...
 	values := make([][]byte, NodeWidth)
-	values[k[31]] = value
-	return n.insertMultiple(k[:31], values)
+	values[key[StemSize]] = value
+	return n.insertMultiple(key, values)
```
gballet, line 916: Do you mean 

```suggestion
	return n.insertMultiple(key[:StemSize], values)
```

? You're passing 32 bytes instead of 31, but it is compared to a stem in `insertMultiple`.

config_ipa_test.go
gballet: thanks for sharing, yeah if other packages use this one, we might do that, but as long as no one requests this, there's no point in complicating the code for now.

gballet (commented):
Nice, I left a few comments but LGTM overall.

ipa.go
jsign: Yes, we also have panics in other places which feels uncomfortable for the same reason. 
I think at some point returning an error would mean a complete halt in the geth node since it can't continue with the execution pipeline and is non-recoverable; but definitely a panic is blowing up `defers` and other nicier "good stops" for sure.

I'll create an issue to reconsider all panics in the package, since probably returning errors can have some breaking in geth that maybe we can batch at the same time.

jsign: BTW, we should do the same in `go-ipa`; scan everything looking for panics and handle them differently.  Just to remember.

ipa.go
jsign: Uhm, yeah. Honestly, I'd prefer to keep then `StemFromBytes`, at least add the `len(data) == StemSize` and call `FromLEBytes`.

That sounds better for sure since:
- We include that length check that you mention, and stop relying on callers doing the right thing.
- We avoid the code duplication that was annoying me, since `FromLEBytes` is correct.

I'm going to change to go that route... I'd prefer to be safe as much as possible.

tree.go
```diff
@@ -477,12 +475,12 @@ func (n *InternalNode) GetStem(stem []byte, resolver NodeResolverFn) ([][]byte,
...
 	case *HashedNode:
 		if resolver == nil {
-			return nil, fmt.Errorf("hashed node %x at depth %d along stem %x could not be resolved", child.Commitment().Bytes(), n.depth, stem)
+			return nil, errReadFromInvalid
 		}
 		hash := child.commitment
 		serialized, err := resolver(hash)
 		if err != nil {
-			return nil, fmt.Errorf("verkle tree: error resolving node %x at depth %d: %w", stem, n.depth, err)
+			return nil, err
```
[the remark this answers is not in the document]
jsign: See above comment.

tree.go
```diff
@@ -585,38 +583,21 @@ func (n *InternalNode) FlushAtDepth(depth uint8, flush NodeFlushFn) {
...
-			return nil, err
-		}
-		n.children[nChild] = c
-
-		return c.Get(k, getter)
-	default: // InternalNode
-		return child.Get(k, getter)
 	}
+
+	return stemValues[key[StemSize]], nil
```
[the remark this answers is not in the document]
jsign: Will do.

tree.go
jsign: @gballet, since we're recovering the `MergesTrees(...)` that uses this method, I'm also recovering this one but at least leave it unexported.

tree.go
jsign: `insertMultiple(...)` first parameter is named `key` so passing a key (32 bytes) it should be fine, since the method should handle that semantics. `insertMultiple(...)` calls `equalPaths(...)` which already does the `[:StemSize]` accordingly.

jsign: OK, I see your comments below. If you want to change from `key` to `stem`, then that suggested change is needed. 
Just replied that isn't the case without changing the API.

I can change from `key` to `stem` for sure.

ipa.go
jsign: Done!

jsign:
> Nice, I left a few comments but LGTM overall.

Thanks! I adjusted with some extra commits to many of your comments. I closed all convos except one that I still need to make a decision on; I'll wait for your feedback.

tree.go
```diff
@@ -477,12 +475,12 @@ func (n *InternalNode) GetStem(stem []byte, resolver NodeResolverFn) ([][]byte,
 		return nil, nil
 	case *HashedNode:
 		if resolver == nil {
-			return nil, fmt.Errorf("hashed node %x at depth %d along stem %x could not be resolved", child.Commitment().Bytes(), n.depth, stem)
+			return nil, errReadFromInvalid
```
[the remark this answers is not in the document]
gballet: I'm fine with `%w`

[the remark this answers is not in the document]
jsign: @gballet, [done](https://github.com/gballet/go-verkle/pull/350/commits/8a4880bdc19aa4a8044b3c89ca48ea9baf49a081).

jsign:
@gballet, all ready.

gballet (approved):
LGTM thanks

PR #352, opened 2023-05-08
deleting from a missing node is a noop

gballet:
This is probably not the root cause why the translated replay fails, but deleting a location that does not exist should work as it does in the MPT.


jsign (approved):
LGTM

PR #353, opened 2023-05-11
fix: set nodes known to be absent to Empty, to differentiate from unknown nodes

gballet:
A reconstructed stateless tree would not set its (known) absent values to `Empty` but to `Unknown` which caused `GetProofItems` to error. Set the reconstructed 'absent' tree to `Empty` to fix this problem.

### DIFF
--- a/proof_test.go
+++ b/proof_test.go
@@ -637,8 +637,8 @@ func TestStatelessDeserializeMissginChildNode(t *testing.T) {
 		t.Fatal("differing commitment for child #0")
 	}
 
-	if droot.(*InternalNode).children[64] != UnknownNode(struct{}{}) {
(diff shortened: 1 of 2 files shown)


proof_test.go
```diff
@@ -637,8 +637,8 @@ func TestStatelessDeserializeMissginChildNode(t *testing.T) {
 		t.Fatal("differing commitment for child #0")
 	}
 
-	if droot.(*InternalNode).children[64] != UnknownNode(struct{}{}) {
-		t.Fatalf("non-nil child #64: %v", droot.(*InternalNode).children[64])
```
gballet, line 641: The point of this test is to ensure that the node is missing, but that it is known from the proof that it is missing.

PR #354, opened 2023-05-17
Remove all references to children in InternalNode & use uncompressed serialization

gballet:
Remove all references to children commitments in `InternalNode` so that the uncompressed form can be stored inside the child's node itself, and not be needed by the parent. This supposed a path-based DB schema.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/gballet/go-verkle
 
 go 1.19
 
-require github.com/crate-crypto/go-ipa v0.0.0-20230601170251-1830d0757c80
+require github.com/crate-crypto/go-ipa v0.0.0-20230626131944-6a9b06cf26df
 
 require (
 	golang.org/x/sync v0.1.0 // indirect
--- a/tree.go
+++ b/tree.go
@@ -858,12 +858,18 @@ func (n *InternalNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]
 }
 
 func (n *InternalNode) Serialize() ([]byte, error) {
-	ret := make([]byte, nodeTypeSize+SerializedPointCompressedSize)
+	ret := make([]byte, nodeTypeSize+bitlistSize+SerializedPointCompressedSize)
+	bitlist := ret[internalBitlistOffset:internalNodeChildrenOffset]
+	for i, c := range n.children {
+		if _, ok := c.(Empty); !ok {
+			setBit(bitlist[:], i)
+		}
+	}
 
 	// Store in ret the serialized result
 	ret[nodeTypeOffset] = internalRLPType
 	comm := n.commitment.Bytes()
-	copy(ret[internalBitlistOffset:], comm[:])
+	copy(ret[internalNodeChildrenOffset:], comm[:])
 	// Note that children were already appended in ret through the children slice.
 
 	return ret, nil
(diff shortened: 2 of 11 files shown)


tree.go
```diff
@@ -1405,8 +1381,8 @@ func (n *LeafNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]byte
 // Serialize serializes a LeafNode.
 // The format is: <nodeType><stem><bitlist><c1comm><c2comm><children...>
 func (n *LeafNode) Serialize() ([]byte, error) {
-	cBytes := banderwagon.ElementsToBytes([]*banderwagon.Element{n.c1, n.c2})
-	return n.serializeWithCompressedCommitments(cBytes[0], cBytes[1]), nil
+	cBytes := banderwagon.ElementsToBytes([]*banderwagon.Element{n.commitment, n.c1, n.c2})
+	return n.serializeLeafWithCompressedCommitments(cBytes[0], cBytes[1], cBytes[2]), nil
```
jsign, line 1385: You mentioned in the PR description that:
> .. so that the uncompressed form can be stored inside the child's node itself

Did you meant "compressed" form? Or should we change the code? Or maybe is deferred?
(Uncompressed is attractive to avoid costly deserialization later, which I think was part of the idea (apart from saving space))

tree_test.go
```diff
@@ -592,45 +591,42 @@ func TestNodeSerde(t *testing.T) {
...
 		t.Error(err)
 	}
-	rc := root.commitment.Bytes()
 
 	// Now deserialize and re-construct tree
-	res, err := ParseNode(ls0, 1, l0c[:])
+	res, err := ParseNode(ls0, 1)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
```
jsign, line 609: Yeah, I prefer earlier.

jsign (approved):
Very interesting to see how this plays out!
I left some comments to get your thoughts on some parts.

tree.go
gballet: the goal is to have the uncompressed form, but right now I just want to see the feasibility of path-based, to compare the performance - then if we can improve by not serializing points, that's even better - but let's see.

tree.go
```diff
@@ -338,24 +338,23 @@ func (n *InternalNode) Insert(key []byte, value []byte, resolver NodeResolverFn)
...
+		n.cowChild(nChild)
 		n.children[nChild] = NewLeafNode(stem, values)
 		n.children[nChild].setDepth(n.depth + 1)
 	case *HashedNode:
 		if resolver == nil {
 			return errInsertIntoHash
 		}
-		hash := child.commitment
-		serialized, err := resolver(hash)
+		serialized, err := resolver(stem[:n.depth+1])
```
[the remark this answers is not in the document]
gballet: nice catch, was debugging and just found out that was a problem :|

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/gballet/go-verkle
 
 go 1.18
 
-require github.com/crate-crypto/go-ipa v0.0.0-20230601170251-1830d0757c80
+require github.com/crate-crypto/go-ipa v0.0.0-20230626131944-6a9b06cf26df
```
jsign, line 5: Reminder before merging: this is a tentative PR in go-ipa. Before pushing to `master`, I should open a PR there to merge it in go-ipa first.

jsign: [Reference](https://github.com/gballet/go-verkle/pull/370)

PR #356, opened 2023-05-18
TreeFromProof test hardening

jsign:
This PR adds more test coverage for `PreStateTreeFromProof(...)`.

The diff is very messy, so I recommend some "split view" maybe.

### DIFF
--- a/proof_test.go
+++ b/proof_test.go
@@ -738,7 +738,7 @@ func TestStatelessDeserialize(t *testing.T) {
 	testSerializeDeserializeProof(t, insertKVs, proveKeys)
 }
 
-func TestStatelessDeserializeMissingChildNode(t *testing.T) {
+func TestStatelessDeserializeMissingLeafNode(t *testing.T) {
 	t.Parallel()
 
 	insertKVs := map[string][]byte{
@@ -747,6 +747,18 @@ func TestStatelessDeserializeMissingChildNode(t *testing.T) {
 		string(ffx32KeyTest): fourtyKeyTest,
 	}
 	proveKeys := keylist{zeroKeyTest, fourtyKeyTest}
+
+	testSerializeDeserializeProof(t, insertKVs, proveKeys)
+}
+
+func TestStatelessDeserializeAbsentValueInExistingLeafNode(t *testing.T) {
+	t.Parallel()
+
+	insertKVs := map[string][]byte{
+		string(zeroKeyTest):  fourtyKeyTest,
+		string(ffx32KeyTest): fourtyKeyTest,
+	}
+	proveKeys := keylist{zeroKeyTest, oneKeyTest}
 
 	testSerializeDeserializeProof(t, insertKVs, proveKeys)
 }
@@ -804,10 +816,12 @@ func testSerializeDeserializeProof(t *testing.T, insertKVs map[string][]byte, pr
 	root := New()
 
 	for k, v := range insertKVs {
-		root.Insert([]byte(k), v, nil)
+		if err := root.Insert([]byte(k), v, nil); err != nil {
+			t.Fatalf("could not insert key: %v", err)
+		}
 	}
+	root.Commit()
 
-	// TODO(jsign): MakeVerkleMultiProof has a strange-ish APIs, reconsider this when digging into why that's the case.
 	absentKeys := map[string]struct{}{}
 	proveKVs := map[string][]byte{}
 	for _, key := range proveKeys {
--- a/proof_test.go
+++ b/proof_test.go
@@ -32,8 +32,6 @@ import (
 	"fmt"
 	"reflect"
 	"testing"
-
-	"github.com/crate-crypto/go-ipa/common"
 )
 
 func TestProofEmptyTree(t *testing.T) {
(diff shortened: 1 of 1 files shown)


proof_ipa.go
```diff
@@ -353,6 +354,7 @@ func TreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) {
 				continue
 			}
 
+			// TODO(jsign): this if will never be true.
 			if bytes.Equal(k, info[string(p)].stem) {
 				values[k[31]] = proof.Values[i]
 			}
```
jsign, line 360: @gballet, doesn't seem that this `if` will ever be true. We're comparing a `k` (32 byte key) with a stem which is 31 byte.

proof_ipa.go
```diff
@@ -79,6 +79,7 @@ func GetCommitmentsForMultiproof(root VerkleNode, keys [][]byte) (*ProofElements
 	return root.GetProofItems(keylist(keys))
 }
 
+// QUESTION(jsign): do we need this `keys` and `keyvals` separation? Shouldn't keys == keys(keyvals)?
 func MakeVerkleMultiProof(root VerkleNode, keys [][]byte, keyvals map[string][]byte) (*Proof, []*Point, []byte, []*Fr, error) {
```
jsign, line 83: @gballet, why we'd want a `keyvals` map that contains key-values that aren't in `keys`?
The method seems to ignore keys in `keyvals` that aren't in `keys`.

Also, in L103 there's an assumption that `keyvals[string(k)]` will return a value; but technically it could return a default value if the key isn't present in the map. Feels like an obscure assumption too.

tree_test.go
```diff
@@ -1178,6 +1178,7 @@ func TestEmptyHashCodeCachedPoint(t *testing.T) {
 }
 
 func TestBatchMigratedKeyValues(t *testing.T) {
+	t.SkipNow()
```
jsign, line 1181: TODO, remove this line. (Actually, I want to remove this slow test after we merge https://github.com/gballet/go-verkle/pull/349)

proof_test.go
```diff
@@ -571,106 +572,162 @@ func TestVerkleProofMarshalUnmarshalJSON(t *testing.T) {
...
 
-func TestStatelessDeserializeMissginChildNode(t *testing.T) {
-	root := New()
-	for _, k := range [][]byte{zeroKeyTest, oneKeyTest, ffx32KeyTest} {
-		root.Insert(k, fourtyKeyTest, nil)
-	}
-
-	proof, _, _, _, _ := MakeVerkleMultiProof(root, keylist{zeroKeyTest, fourtyKeyTest}, map[string][]byte{string(zeroKeyTest): fourtyKeyTest, string(fourtyKeyTest): nil})
+func TestProofVerificationThreeStemsInSameExtensionStatus(t *testing.T) {
+	key1_0, _ := hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000000")
```
gballet, line 610: that's `zeroKeyTest`

proof_test.go
```diff
@@ -571,106 +572,162 @@ func TestVerkleProofMarshalUnmarshalJSON(t *testing.T) {
...
-	}
+	testSerializeDeserializeProof(t, insertKVs, proveKeys)
+}
 
-	droot, err := TreeFromProof(dproof, root.Commit())
-	if err != nil {
-		t.Fatal(err)
-	}
+func TestProofVerificationTwoLeavesWithDifferentValues(t *testing.T) {
+	key1, _ := hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000000")
```
gballet, line 626: that's `zeroKeyTest`

proof_test.go
```diff
@@ -571,106 +572,162 @@ func TestVerkleProofMarshalUnmarshalJSON(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("error deserializing proof: %v", err)
-	}
+	testSerializeDeserializeProof(t, insertKVs, proveKeys)
+}
 
-	droot, err := TreeFromProof(dproof, root.Commit())
-	if err != nil {
-		t.Fatal(err)
+func TestStatelessDeserializeMissingChildNode(t *testing.T) {
```
gballet, line 586: that's nice, but there should be another one in which `oneKeyTest` is the missing key, so that we can check this use case is also working.

gballet (commented):
Thanks for looking at that. Indeed a lot of stuff has fossilized and should be cleaned up. I will change the interface to `MakeVerkleMultiProof`

gballet:
@jsign right, the reason `keyvals` is passed, is because I used to make the difference between "previous" values (held by `keyvals`) and "current values" (held by the tree). This isn't something that we are doing at the moment, but it will come back since we want to put the post state in the tree. We can figure out the proper way to do this when this happens.

proof_test.go
```diff
@@ -988,87 +1074,6 @@ func TestProofVerificationWithPostState(t *testing.T) { // skipcq: GO-R1005
 		})
 	}
 }
-func TestProofOfAbsenceBorderCase(t *testing.T) {
```
jsign, line 991: Not deleted, it's now in L813.

proof_test.go
```diff
@@ -988,87 +1074,6 @@ func TestProofVerificationWithPostState(t *testing.T) { // skipcq: GO-R1005
...
-	if !droot.Commit().Equal(root.Commit()) {
-		t.Fatal("differing root commitments")
-	}
-
-	if !droot.(*InternalNode).children[0].Commit().Equal(root.(*InternalNode).children[0].Commit()) {
-		t.Fatal("differing commitment for child #0")
-	}
-}
-
-func TestProofOfAbsenceBorderCaseReversed(t *testing.T) {
```
jsign, line 1032: Not deleted, it's no in L826.

proof_test.go
jsign: Done, it's named `TestStatelessDeserializeAbsentValueInExistingLeafNode(...)`.

PR #357, opened 2023-05-18
implement proper node deletion in tree

gballet:
The PBSS scheme uses reverse-diffs to handle reorgs past the persistent disk layer. For this to work with verkle, `Delete` must be able to completely delete the nodes that were inserted from one tree to the next.

TODO

 * [x] complete leaf deletion to correctly update the commitments
 * [x] fix existing tests
 * [x] add tests to ensure children corner cases are covered
 * [x] swap `delete` and `err`

### DIFF
--- a/empty.go
+++ b/empty.go
@@ -35,8 +35,8 @@ func (Empty) Insert([]byte, []byte, NodeResolverFn) error {
 	return errDirectInsertIntoEmptyNode
 }
 
-func (Empty) Delete([]byte, NodeResolverFn) error {
-	return errors.New("cant delete an empty node")
+func (Empty) Delete([]byte, NodeResolverFn) (error, bool) {
+	return errors.New("cant delete an empty node"), false
 }
 
 func (Empty) Get([]byte, NodeResolverFn) ([]byte, error) {
--- a/empty.go
+++ b/empty.go
@@ -35,8 +35,8 @@ func (Empty) Insert([]byte, []byte, NodeResolverFn) error {
 	return errDirectInsertIntoEmptyNode
 }
 
-func (Empty) Delete([]byte, NodeResolverFn) (error, bool) {
-	return errors.New("cant delete an empty node"), false
+func (Empty) Delete([]byte, NodeResolverFn) (bool, error) {
+	return false, errors.New("cant delete an empty node")
 }
 
 func (Empty) Get([]byte, NodeResolverFn) ([]byte, error) {
--- a/tree.go
+++ b/tree.go
@@ -1167,7 +1167,11 @@ func (n *LeafNode) Delete(k []byte, _ NodeResolverFn) (bool, error) {
 		n.commitment.Sub(n.commitment, cfg.CommitToPoly(poly[:], 0))
 
 		// Clear the corresponding commitment
-		cn.SetBytes(nil)
+		if k[31] < 128 {
+			n.c1 = nil
+		} else {
+			n.c2 = nil
+		}
 
 		return false, nil
 	}
--- a/tree.go
+++ b/tree.go
@@ -1145,17 +1145,35 @@ func (n *LeafNode) Delete(k []byte, _ NodeResolverFn) (bool, error) {
 	// of removing the last value, update C by
 	// adding -Cn to it and exit.
 	if isCnempty {
-		var cn *Point
+		var (
+			cn           *Point
+			subtreeindex = 2 + k[31]/128
+		)
+
 		if k[31] < 128 {
 			cn = n.c1
 		} else {
 			cn = n.c2
 		}
-		n.commitment.Add(n.commitment, cn.Neg(cn))
+
+		// Update C by subtracting the old value for Cn
+		// Note: this isn't done in one swoop, which would make sense
+		// since presumably a lot of values would be deleted at the same
+		// time when reorging. Nonetheless, a reorg is an already complex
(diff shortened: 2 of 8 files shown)


empty.go
```diff
@@ -35,8 +35,8 @@ func (Empty) Insert([]byte, []byte, NodeResolverFn) error {
 	return errDirectInsertIntoEmptyNode
 }
 
-func (Empty) Delete([]byte, NodeResolverFn) error {
-	return errors.New("cant delete an empty node")
+func (Empty) Delete([]byte, NodeResolverFn) (error, bool) {
```
jsign, line 38: The return order is pretty weird (non-idiomatic), shouldn't be `(bool, error)`?

tree.go
```diff
@@ -513,30 +513,52 @@ func (n *InternalNode) toHashedNode() *HashedNode {
...
 	default:
 		n.cowChild(nChild)
-		return child.Delete(key, resolver)
+		err, delete := child.Delete(key, resolver)
+		if err != nil {
+			return err, false
+		}
+
+		// delete the entire child if instructed to by
+		// the recursive algorightm.
```
jsign, line 545: Typo.

tree.go
```diff
@@ -513,30 +513,52 @@ func (n *InternalNode) toHashedNode() *HashedNode {
...
+		if delete {
+			n.children[nChild] = Empty{}
```
jsign, line 547: This will be an obvious comment, but we're changing the spec, right?

tree.go
```diff
@@ -1075,15 +1097,76 @@ func (n *LeafNode) updateMultipleLeaves(values [][]byte) {
...
+				isCnempty = false
+				isCempty = false
+				break
+			} else {
+				// i and k[31] were in a different subtree,
+				// so all we can say at this stage, is that
+				// the whole tree isn't empty.
+				// TODO if i < 128, then k[31] >= 128 and
+				// we could skip to 128, but that's an
+				// optimization for later.
```
jsign, line 1132: Yes, or be a bit smarter in L1118 (for initialization) to only iterate in the right subtree.
e.g: `i:=k[31]/12*128; i<k[31]/12*128+128; i++` ish
Prob requires some extra `C1 vs C2 != empty` checks to fully complete the optimization.

tree.go
```diff
@@ -1075,15 +1097,76 @@ func (n *LeafNode) updateMultipleLeaves(values [][]byte) {
...
+	// This is done by setting the leaf value
+	// to `nil` at this point, and all the
+	// diff computation will be performed by
+	// updateLeaf since leafToComms supports
+	// nil values.
+	// Note that the value is set to nil by
+	// the method, as it needs the original
+	// value to compute the commitment diffs.
+	n.values[k[31]] = original
+	n.updateLeaf(k[31], nil)
```
jsign, line 1168: Doesn't `updateLeaf(...)` updates commitments by using the deletion-marker?
That seems to be a deletion semantics different from all the above logic in this method, which "completely wipes out" the existence of the original value in the commitment.

Looks weird, but I could be missing context. Mentioning it just in case.

jsign (commented):
Left some comments.

empty.go
jsign: NVM, I see your todo `swap delete and err`.

tree.go
gballet: we're not changing the spec in the sense that going forward along the chain, no deletion is possible. But in the case of reorgs, it must be possible to delete values that were inserted in the tree while executing the reorg'ed blocks.

tree.go
gballet: Ah yes good point, I could indeed check that `Cn != zero`, and then only go over one subtree :+1:

tree.go
gballet: So to be completely clear here: instead of calling `Delete(key)` in `(*VerkleTrie).Delete`, one should call `Update(key, 0)` to overwrite it with 0s. Geth is responsible for following the spec, this library implements proper deletion as it will be used for the path-based approach to implement deep reorgs.

tree.go
gballet: But the Cn calulcation in the other case was clearly wrong.

tree.go
gballet: not going to fix that today though

tree.go
jsign: Ah, OK, so it's an API usage change from the client side compared to before. Makes sense.

tree.go
```diff
@@ -1145,17 +1145,35 @@ func (n *LeafNode) Delete(k []byte, _ NodeResolverFn) (bool, error) {
...
+		// since presumably a lot of values would be deleted at the same
+		// time when reorging. Nonetheless, a reorg is an already complex
+		// operation which is slow no matter what, so ensuring correctness
+		// is more important than
+		var poly [4]Fr
+		toFr(&poly[subtreeindex], cn)
+		n.commitment.Sub(n.commitment, cfg.CommitToPoly(poly[:], 0))
+
+		// Clear the corresponding commitment
+		cn.SetBytes(nil)
```
jsign, line 1170: In `tree.go` we usually do `n.c1 != nil` style to detect empty commitments, so this `cn.SetBytes(nil)` is not leaving things in that "style". I I think we might need to set `nil` instead of "clear out" the content. 

So that means falling back to an `if` statement similar to L1153, since we can't do that with `cn` indirection.

jsign (approved):

PR #358, opened 2023-05-26
do not error when deleting non-existing leaves

gballet:
Because of the way the overlay transition has been coded, it can happen that an attempt is made to delete from the overlay tree. This occurs when an account has not been fully translated, and that the calling code will try to delete one of its values. The code will first try to delete the data from the overlay tree, and then from the MPT (which isn't allowed).

:thinking: come to think of it, this is probably the wrong thing to do: a zero value should be inserted instead, so that the MPT is never hit and that there is a record that the value got deleted.  I will attempt to fix it in a different PR, but in the meantime this PR fixes the replay issue.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -432,7 +432,7 @@ func (n *InternalNode) Delete(key []byte, resolver NodeResolverFn) error {
 	nChild := offset2key(key, n.depth)
 	switch child := n.children[nChild].(type) {
 	case Empty:
-		return errDeleteNonExistent
+		return nil
 	case *HashedNode:
 		if resolver == nil {
 			return errDeleteHash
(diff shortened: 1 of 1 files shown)


jsign:
> thinking come to think of it, this is probably the wrong thing to do: a zero value should be inserted instead, so that the MPT is never hit and that there is a record that the value got deleted. I will attempt to fix it in a different PR, but in the meantime this PR fixes the replay issue.

Yep, if we don't do this, if the MPT slot has a value the conversion will migrate it to VKT which would be incorrect. (i.e: will copy a stale value)

jsign (approved):
Leaving it approved if we want to accept this temporarily (I'll leave that up to you).
(Not sure if asking for review was intentional actually).

gballet:
I implemented the overwriting of leaves with zeroes, which is the proper fix. Closing.

PR #362, opened 2023-06-07
tree: allow batching new leaf nodes

gballet:
See #349, looks like github got confused by some forced push.


gballet:
@jsign it was for an investigation, I swear! #363, however, is totally meant to annoy you 😁 Closing this one.

PR #363, opened 2023-06-07
tree: allow batching new leaf nodes

gballet:
Try again to run #349 with a different branch name...

### DIFF
--- a/tree_test.go
+++ b/tree_test.go
@@ -1323,7 +1323,9 @@ func TestManipulateChildren(t *testing.T) {
 
 func TestLeafNodeInsert(t *testing.T) {
 	valIdx := 42
-	ffx31plus42 := append(ffx32KeyTest[:StemSize], byte(valIdx))
+	ffx31plus42 := make([]byte, 32)
+	copy(ffx31plus42, ffx32KeyTest[:StemSize])
+	ffx31plus42[StemSize] = byte(valIdx)
 
 	tree := New()
 	if err := tree.Insert(ffx31plus42, testValue, nil); err != nil {


tree.go
```diff
@@ -248,47 +251,14 @@ func NewStatelessInternal(depth byte, comm *Point) VerkleNode {
...
-
 	return &LeafNode{
 		// depth will be 0, but the commitment calculation
 		// does not need it, and so it won't be free.
 		values:     values,
-		stem:       stem,
-		commitment: cfg.CommitToPoly(poly[:], NodeWidth-4),
-		c1:         c1,
-		c2:         c2,
+		stem:       stem[:StemSize],
```
gballet, line 258: This was the issue: your test is checking that `Serialize` would trim the stem up to `StemSize`, but it would call `Commit` first. Now, there is a check inside commit to ensure that the stem size is correct. This was introduced recently, so this is why this test (which didn't exist in `jsign/cowleaves2`) didn't fail.

gballet (commented):
@jsign ok so it was weird, but it wasn't really github action's fault: CI will merge your PR branch with main during the test, making sure that the merge can work. It's something that needs to happen.

The problem is that if you rebase your branch on top of `master`, there are a few problems appearing. This is why it worked locally for both of us: we were not rebasing. I have fixed the first problem, but there is now one with a `nil` commitment.

tree_test.go
```diff
@@ -1322,13 +1322,20 @@ func TestManipulateChildren(t *testing.T) {
 }
 
 func TestLeafNodeInsert(t *testing.T) {
-	values := make([][]byte, NodeWidth)
 	valIdx := 42
-	values[valIdx] = testValue
-	ln := NewLeafNode(ffx32KeyTest[:StemSize], values)
```
jsign, line 1391: We can't create "direct leaf node" now since all the internal data is lazily computed. This caused a crash later when doing `ln.Hash()` called `(*LeafNode).Commitment()`, which panicked since there wasn't any calculated commitment.

The solution is always to use the tree from a properly created tree, since before inspecting internals of things you should always call `tree.Commitment()` to do all the pending calculations.

Another solution is making `(*LeafNode).Commitment()` not panicked if `commitment == nil`, so compute it if that's the case. Technically speaking, that would be my preferred solution since we shouldn't have panicked but considering that case should never happen...

gballet (commented):
thanks for fixing this, I rebased it and will merge it after we discuss your comment.

gballet:
Note to self when I revisit this: This PR is kept on hold for a little bit, as it needs to be rebased and it fill conflict with more upcoming PRs. It should not decrease performance, so it's safe to merge.

jsign:
> Note to self when I revisit this: This PR is kept on hold for a little bit, as it needs to be rebased and it fill conflict with more upcoming PRs. It should not decrease performance, so it's safe to merge.

Update: rebased to `master`. Still agree we can keep it on hold, but hopefully will be less painful to keep rebasing if we keep merging more stuff in `master`.

Issue #351, opened 2023-05-02
Avoid panics


jsign:
Do a full-scan of the repository to avoid explicit panics. Even though we usually use them in cases where critical invariants must hold, panicking is very aggressive for at least a graceful shutdown, even in non-recoverable errors.

Also, scan for dependencies (e.g: `go-ipa`) for the same.

We can do this getting closer to a stable implementation, and definitely before getting close to being fully integrated in geth since it can result in breaking some APIs returning errors instead of panicking.

cc @gballet

gballet:
Yeah that's too many

```
~/s/go-verkle (master)> grep -r panic * | wc -l
29
```

PR #366, opened 2023-06-16
remove panics in StemFromBytes and FromLEBytes

gballet:
This is part (but not all) of the fix for #351, stopping here because the diff is already quite large.

### DIFF
--- a/config_ipa.go
+++ b/config_ipa.go
@@ -85,7 +85,9 @@ func GetConfig() *Config {
 		values := make([][]byte, NodeWidth)
 		values[CodeHashVectorPosition] = emptyHashCode
 		var c1poly [NodeWidth]Fr
-		fillSuffixTreePoly(c1poly[:], values[:NodeWidth/2])
+		if _, err := fillSuffixTreePoly(c1poly[:], values[:NodeWidth/2]); err != nil {
+			panic(err)
+		}
 		EmptyCodeHashPoint = *cfg.CommitToPoly(c1poly[:], 0)
 		EmptyCodeHashFirstHalfValue = c1poly[EmptyCodeHashFirstHalfIdx]
 		EmptyCodeHashSecondHalfValue = c1poly[EmptyCodeHashSecondHalfIdx]
--- a/tree.go
+++ b/tree.go
@@ -247,13 +247,16 @@ func NewStatelessInternal(depth byte, comm *Point) VerkleNode {
 }
 
 // New creates a new leaf node
-func NewLeafNode(stem []byte, values [][]byte) *LeafNode {
+func NewLeafNode(stem []byte, values [][]byte) (*LeafNode, error) {
 	cfg := GetConfig()
 
 	// C1.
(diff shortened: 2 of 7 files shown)


tree.go
```diff
@@ -1230,41 +1259,48 @@ func (n *LeafNode) Commit() *Point {
...
 	}
 	copy(valLoWithMarker[:loEnd], val[:loEnd])
 	valLoWithMarker[16] = 1 // 2**128
-	FromLEBytes(&poly[0], valLoWithMarker[:])
+	if err = FromLEBytes(&poly[0], valLoWithMarker[:]); err != nil {
+		return err
+	}
 	if len(val) >= 16 {
-		FromLEBytes(&poly[1], val[16:])
+		err = FromLEBytes(&poly[1], val[16:])
```
jsign, line 1301: Nit: despite this is correct, my paranoid side forces me to warn about: "if tomorrow somebody adds some extra code between lines 1301 and 1303, this can become a bug". 

So maybe simply do:
```
  ...
  if err := FromLEBytes(..., ...); err != nil { 
      return fmt.Errorf("converting to bytes: %s", err)
  }

   return nil
}
```

tree.go
```diff
@@ -1285,7 +1321,10 @@ func (n *LeafNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]byte
 
 	// Initialize the top-level polynomial with 1 + stem + C1 + C2
 	poly[0].SetUint64(1)
-	StemFromBytes(&poly[1], n.stem)
+	err := StemFromBytes(&poly[1], n.stem)
+	if err != nil {
```
jsign, line 1325: Nit: inline err-if?

tree_test.go
```diff
@@ -858,12 +858,12 @@ func TestEmptyCommitment(t *testing.T) {
...
-		if r := recover(); r == nil {
-			t.Fatal("didn't catch length error")
-		}
 	}()
 	var value [33]byte
-	leafToComms([]Fr{}, value[:])
+	err := leafToComms([]Fr{}, value[:])
+	if err == nil {
+		t.Fatal("didn't catch length error")
+	}
```
jsign, line 866: Cool :)

jsign (approved):
LGTM, left some comments for your consideration since I think we require some nit extra changes.

config_ipa.go
```diff
@@ -85,7 +85,9 @@ func GetConfig() *Config {
 		values := make([][]byte, NodeWidth)
 		values[CodeHashVectorPosition] = emptyHashCode
 		var c1poly [NodeWidth]Fr
-		fillSuffixTreePoly(c1poly[:], values[:NodeWidth/2])
+		if _, err := fillSuffixTreePoly(c1poly[:], values[:NodeWidth/2]); err != nil {
+			panic(err)
+		}
```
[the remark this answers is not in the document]
gballet: nah, because if you get an error at this stage, there's no point in continuing (and also it would make a huge diff as it's called everywhere)

PR #367, opened 2023-06-16
conversion: use errgroup to handle failing routines

jsign:
Use [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup), which helps spin up goroutines that can fail and [abort as soon as the first one fails and catch the error](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go) without race conditions.

Context: [See this comment](https://github.com/gballet/go-verkle/pull/366#discussion_r1232151007).

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -2,10 +2,12 @@ package verkle
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"runtime"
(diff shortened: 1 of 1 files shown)


gballet (approved):
LGTM, thanks for catching that.

PR #369, opened 2023-06-22
Remove needing a file persisting precomputed tables

jsign:
This PR updates *go-ipa* dependency to a new version. This new version has a faster way of precomputing tables for MSM, removing the need to store them in a file.

As a bonus, I also switched all our tests to be run in parallel (looking at CI, it changed from ~13min to ~5min) 
I also found a small bug in one of them since it mutated a shared key. I’ll touch on this in a comment. 

For more details see: https://github.com/crate-crypto/go-ipa/pull/51


.github/workflows/block_replay.yml
```diff
@@ -26,7 +26,6 @@ jobs:
     - name: Edit geth to use this repo and build it
```
jsign, line 26: Removing `precomp` related actions.
I think we could remove the already existing file in the CI machine? Not a rush to do it, but it shouldn't be needed now.

config_ipa.go
```diff
@@ -27,7 +27,7 @@ package verkle
 
 import (
```
jsign, line 28: I'd recommend to see this diff in "split" view since the default view can be confusing.

config_ipa.go
```diff
@@ -46,41 +46,34 @@ const (
...
+	onceCfg sync.Once
+)
 
-func (ipac *IPAConfig) CommitToPoly(poly []Fr, _ int) *Point {
-	ret := ipac.conf.Commit(poly)
-	return &ret
+func init() {
+	FrZero.SetZero()
+	FrOne.SetOne()
 }
```
jsign, line 60: Just organizing a bit all const/vars on the top of the file.

config_ipa.go
```diff
@@ -46,41 +46,34 @@ const (
...
-			} else if err = os.WriteFile(precompFileName, serialized, 0666); err != nil {
-				panic("error saving the precomp: " + err.Error())
-			}
-		} else {
-			srs, err := ipa.DeserializeSRSPrecomp(precompSer)
-			if err != nil {
-				panic("error deserializing precomputed Lagrange points:" + err.Error())
-			}
-			ipacfg = ipa.NewIPASettingsWithSRSPrecomp(srs)
+	onceCfg.Do(func() {
```
jsign, line 69: I prefer to be safe and use `sync.Once` here. Our previous way can be racy and trust the client to not fire two goroutines that would call `GetConfig()` and mess everything up.

It also allows the tests to run in parallel since it doesn't care who calls (and if done in parallel) `GetConfig()`.

config_ipa.go
```diff
@@ -46,41 +46,34 @@ const (
...
+		conf, err := ipa.NewIPASettings()
+		if err != nil {
+			panic(err)
```
jsign, line 72: The previous code panicked, and I was very tempted to allow `GetConfig()` to return an error.
It should be impossible to have an error in the way we call it here. The only way `NewIPASettings()` an error is that if you provide less than 256-points (internally), which is "never".

If we return an `error` in `GetConfig()`, that can be a bit annoying in calling sites that don't return an error will have to do it or also decide to panic.

For now leaving this as a comment since this isn't adding a new panic case. If this panics, it'd be at node startup time.

config_ipa.go
```diff
@@ -89,17 +82,11 @@ func GetConfig() *Config {
...
-	FrZero Fr
-	FrOne  Fr
-)
-
-func init() {
-	FrZero.SetZero()
-	FrOne.SetOne()
+func (conf *IPAConfig) CommitToPoly(poly []Fr, _ int) *Point {
+	ret := conf.conf.Commit(poly)
+	return &ret
```
jsign, line 91: Just moved code, it's the same.

config_ipa_test.go
jsign, line 1: We don't need "file" tests anymore.

debug_test.go
```diff
@@ -30,6 +30,8 @@ import (
 )
 
 func TestJSON(t *testing.T) {
+	t.Parallel()
```
jsign, line 33: Parallelizing everything. [There will be an extra comment in this PR for a bugfix in a test]

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/gballet/go-verkle
 
 go 1.19
 
-require github.com/crate-crypto/go-ipa v0.0.0-20230601170251-1830d0757c80
+require github.com/crate-crypto/go-ipa v0.0.0-20230622122819-a63eaa2f7c20
```
jsign, line 5: Using tenative https://github.com/crate-crypto/go-ipa/pull/51
I'll change when it gets merged.

tree_test.go
```diff
@@ -814,27 +854,29 @@ func TestToDot(*testing.T) {
 
 	// ensure the ToDot output contains a single HashedNode
 	if !strings.Contains(ToDot(root), "hash00") {
-		panic("ToDot output does not contain a HashedNode")
+		t.Error("ToDot output does not contain a HashedNode")
```
jsign, line 857: Avoid panics in tests.

tree_test.go
```diff
@@ -1209,78 +1271,85 @@ func TestEmptyHashCodeCachedPoint(t *testing.T) {
...
-				// ***Insert the key pairs 'naively' ***
-				rand := mRand.New(mRand.NewSource(42)) //skipcq: GSC-G404
-				tree := genRandomTree(rand, treeInitialKeyValCount)
-				randomKeyValues := genRandomKeyValues(rand, migrationKeyValueCount)
-
-				now := time.Now()
-				for _, kv := range randomKeyValues {
-					if err := tree.Insert(kv.key, kv.value, nil); err != nil {
-						t.Fatalf("failed to insert key: %v", err)
+		treeInitialKeyValCount := treeInitialKeyValCount
```
jsign, line 1279: Isn't worth reviewing much, but I'm just parallelizing this test.

tree_test.go
```diff
@@ -1385,13 +1456,17 @@ func TestManipulateChildren(t *testing.T) {
...
+	copy(keyTest, ffx32KeyTest)
 	values := make([][]byte, NodeWidth)
 	valIdx := 42
 	values[valIdx] = testValue
-	ln := NewLeafNode(ffx32KeyTest[:StemSize], values)
+	ln := NewLeafNode(keyTest[:StemSize], values)
 
 	// Check we get the value correctly via Get(...).
-	getValue, err := ln.Get(append(ffx32KeyTest[:StemSize], byte(valIdx)), nil)
+	getValue, err := ln.Get(append(keyTest[:StemSize], byte(valIdx)), nil)
```
jsign, line 1469: Here was the bug that became apparent when paralellizing (quite dangerous!).

The bug was in the deleted line: `append(ffx32KeyTest[:StemSize], byte(valIdx))`
That is literally appending bytes to `ffx32Keytest`, which is global. Now just make a copy in the test and do whatever we want.

jsign:
@gballet, before opening for review, I'll test this myself in the pbss replay bench now that I have the data; just to further confirm all is fine.

config_ipa_test.go
jsign, line 1: We don't need "file" tests anymore.

tree_test.go
```diff
@@ -1391,13 +1466,20 @@ func TestManipulateChildren(t *testing.T) {
...
+	keyTest := make([]byte, len(ffx32KeyTest))
+	copy(keyTest, ffx32KeyTest)
 	values := make([][]byte, NodeWidth)
 	valIdx := 42
 	values[valIdx] = testValue
-	ln, _ := NewLeafNode(ffx32KeyTest[:StemSize], values)
+	ln, err := NewLeafNode(keyTest[:StemSize], values)
+	if err != nil {
+		t.Fatalf("failed to create leaf node: %v", err)
+	}
```
jsign, line 1479: Here was the bug that became apparent when paralellizing (quite dangerous!).

The bug was in the deleted line: `append(ffx32KeyTest[:StemSize], byte(valIdx))`
That is literally appending bytes to `ffx32Keytest`, which is global. Now just make a copy in the test and do whatever we want.

gballet (approved):
LGTM. It was still in review, can I merge it?

jsign:
I've run this again in our overlay tree run, and state-roots are the same for at least the first 800 blocks; so keeps looking good.
Pinged Kev, and it already review and merged.

I've updated `go.mod` to point to the new `go-ipa@master`, so all is ready here.

PR #371, opened 2023-06-29
tree: avoid allocs on paths collection

jsign:
This PR avoids a big source of allocations detected in our overlay transition benchmark.

I'll explain in the comments.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -1542,14 +1542,14 @@ func (n *InternalNode) collectNonHashedNodes(list []VerkleNode, paths [][]byte, 
 	list = append(list, n)
 	paths = append(paths, path)
 	for i, child := range n.children {
-		childpath := make([]byte, len(path)+1)
-		copy(childpath, path)
-		childpath[len(path)] = byte(i)
 		switch childNode := child.(type) {
 		case *LeafNode:
 			list = append(list, childNode)
-			paths = append(paths, childpath)
+			paths = append(paths, childNode.stem[:len(path)+1])
 		case *InternalNode:
+			childpath := make([]byte, len(path)+1)
+			copy(childpath, path)
+			childpath[len(path)] = byte(i)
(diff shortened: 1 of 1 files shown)


tree.go
```diff
@@ -1542,14 +1542,14 @@ func (n *InternalNode) collectNonHashedNodes(list []VerkleNode, paths [][]byte,
 	list = append(list, n)
 	paths = append(paths, path)
 	for i, child := range n.children {
-		childpath := make([]byte, len(path)+1)
-		copy(childpath, path)
-		childpath[len(path)] = byte(i)
```
jsign, line 1547: This was unfortunately creating the L1545 slice for every 256 children on every `InternalNode`, even if those children's where empty (or any other case not used in the switch below).

tree.go
```diff
@@ -1542,14 +1542,14 @@ func (n *InternalNode) collectNonHashedNodes(list []VerkleNode, paths [][]byte,
...
 		switch childNode := child.(type) {
 		case *LeafNode:
 			list = append(list, childNode)
-			paths = append(paths, childpath)
+			paths = append(paths, childNode.stem[:len(path)+1])
```
jsign, line 1548: In the case of `LeafNodes` we can reslice the internal `stem`, so directly avoid allocations at this level.

tree.go
```diff
@@ -1542,14 +1542,14 @@ func (n *InternalNode) collectNonHashedNodes(list []VerkleNode, paths [][]byte,
...
 		case *InternalNode:
+			childpath := make([]byte, len(path)+1)
+			copy(childpath, path)
+			childpath[len(path)] = byte(i)
 			list, paths = childNode.collectNonHashedNodes(list, paths, childpath)
```
jsign, line 1553: We only do allocations now in confirmed children that are `InternalNodes`.

gballet (approved):
simple but nasty. Good catch.

PR #372, opened 2023-07-04
tree: transform first layer to hashed nodes after serializing

jsign:
This PR makes `BatchSerialize(...)` transform the first layer of Internal/Leaf nodes to HashedNodes.
This prevents further commitments to serialize the whole tree again. This has a big impact on our overlay-tree benchmarks.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -1538,6 +1538,14 @@ func (n *InternalNode) BatchSerialize() ([]SerializedNode, error) {
 		}
 	}
 
+	for i := range n.children {
+		if ch, ok := n.children[i].(*InternalNode); ok {
+			n.children[i] = ch.toHashedNode()
+		} else if ch, ok := n.children[i].(*LeafNode); ok {
+			n.children[i] = ch.ToHashedNode()
+		}
+	}
+
 	return ret, nil
 }
 
--- a/tree.go
+++ b/tree.go
@@ -1538,6 +1538,10 @@ func (n *InternalNode) BatchSerialize() ([]SerializedNode, error) {
 		}
 	}
 
+	// TODO: we transform nodes in the first layer to HashedNodes, to avoid further calls
+	//       to this method to do double-work. This is a temporary change for geth since in
+	//       the current influx PBSS effort, there're still calls to Commit() storage tries
+	//       which in VKT doesn't make sense anymore. This changes makes those calls a ~noop.
 	for i := range n.children {
 		if ch, ok := n.children[i].(*InternalNode); ok {
 			n.children[i] = ch.toHashedNode()


tree.go
```diff
@@ -1538,6 +1538,14 @@ func (n *InternalNode) BatchSerialize() ([]SerializedNode, error) {
 		}
 	}
 
+	for i := range n.children {
+		if ch, ok := n.children[i].(*InternalNode); ok {
+			n.children[i] = ch.toHashedNode()
+		} else if ch, ok := n.children[i].(*LeafNode); ok {
+			n.children[i] = ch.ToHashedNode()
+		}
+	}
```
jsign, line 1547: Nothing surprising here; this will make further calls to this method to finish fast since all this layer is HashedNodes so it will short-circuit.

gballet (commented):
This makes sense, but I don't think it's the right approach: `(*VerkleTrie).Commit` should simply not be called for storage tries. This is something that won't happen when the state provider Pr is merged.

Still, this change is correct. I'm going to think it over, though, because I wonder if it is really a good idea to replace loaded nodes with their hashes: it might trigger a reload from DB when it was already in RAM a few moments ago.

jsign:
> This makes sense, but I don't think it's the right approach: `(*VerkleTrie).Commit` should simply not be called for storage tries. This is something that won't happen when the state provider Pr is merged.
> 
> Still, this change is correct. I'm going to think it over, though, because I wonder if it is really a good idea to replace loaded nodes with their hashes: it might trigger a reload from DB when it was already in RAM a few moments ago.

Exactly, this isn't the ideal fix until  geth has a proper tree-node caching strategy.

> I'm going to think it over, though, because I wonder if it is really a good idea to replace loaded nodes with their hashes: it might trigger a reload from DB when it was already in RAM a few moments ago.

It isn't a good idea, this change is quite aggressive. If the top X layers are kept serialized in RAM, then reloading them at least wouldn't mean a hit on disk, but it would create memory pressure. Ideally, the top two/three layers _nodes_ shouldn't be destroyed. Maybe other levels could be cached serialized to avoid memory usage.

Definitely, this needs some thought and be aligned with the new `trie.Database` that is being built. As soon that works, I'm pretty sure this change here should be reverted.

gballet:
> Definitely, this needs some thought and be aligned with the new trie.Database that is being built. As soon that works, I'm pretty sure this change here should be reverted.

Could you please add a comment to this effect? Then I will merge it.

jsign:
> > Definitely, this needs some thought and be aligned with the new trie.Database that is being built. As soon that works, I'm pretty sure this change here should be reverted.
> 
> Could you please add a comment to this effect? Then I will merge it.

The `trie.Database` form geth is used to have an intermediate caching layer of nodes before being saved to disk.
That's the right caching layer that should make this PR be reverted at some point. Saying it differently, I think this decision shouldn't be forced at `go-verkle` abstraction layer, but probably geth should be deciding on that (I think `trie.Database`, when it is ready for PBSS, might be that module).

jsign:
Ah, you mean a comment in the code :laughing:  I thought here to clarify.

Yeah, I'll do.

jsign:
@gballet  https://github.com/gballet/go-verkle/pull/372/commits/0fbcda9b6d6bcde53cbb45d40b39c323066d7f6e

gballet (approved):
LGTM

PR #373, opened 2023-07-07
tree: make HashedNode non-pointer and fieldless

jsign:
This PR exploits that in PBSS, HashedNodes can be interpreted as marker nodes of existing nodes not yet loaded in memory. 

The change in this PR makes `HashedNode`:
1. Satisfy the `VerkleNode` interface for `HashedNode` compared to the previous `*HashedNode`.
2. Remove all the existing fields having a similar (equivalent) shape of `Empty` nodes.

These points give us the following benefits (confirmed with benchmarks):
- Avoiding pointers doesn't bias the compiler into assuming these structs escape to the heap.
- Now that the total size of HashedNode is 0 (very small); it triggers a known optimization from the Go compiler that allows to interpret it as `VerkleNode` without forcing a heap allocation since it can embed the value.

Avoiding heap allocations in HeapNode is very important since many HashedNodes exist in a typical in-memory VKT while executing a block. If each existing/unloaded child results in a memory allocation, that adds up.

As a reminder of why making `HashedNode` fieldless works: in PBSS, the `HashedNode` path can be inferred by where it lives in the tree, which is the only thing we know to know the key to ask the `resolver` to load it from disk.

---

The only drawback of this PR is that it makes `FlushAtDepth` API obsolete. Flushing at an arbitrary depth means that `HashedNodes` would need to store the `Commitment` to assist top layers in future flushing. Top layers (not flushed at the selected depth), still have their `cows` marked to be recomputed. Some of their children now `HashedNodes`, and they need to know which is the `Commitment`.  

This isn't a big deal in our current strategy in Geth since we aren't relying on `FlushAtDepth` API in geth for VKTs. This API was useful for the full-tree conversion strategy that is not now the main strategy. We need to confirm that after fully discarding performance problems in low-end ARM machines, which we will probably do soon after doing other work.

This means the repo that some tests that rely on `FlushAtDepth` are temporarily marked as skipped. They were mostly checking if that API was working correctly, so in the end, we'll remove those tests or adjust them to use the full serialization API if they're adding some value to cover any other angle in logic.

I also mention in a PR comment a further tension point of `VerkleNode` and what that means for `HashedNode`.

### DIFF
--- a/hashednode.go
+++ b/hashednode.go
@@ -45,6 +45,12 @@ func (HashedNode) Get([]byte, NodeResolverFn) ([]byte, error) {
 }
 
 func (HashedNode) Commit() *Point {
+	// TODO: we should reconsider what to do with the VerkleNode interface and how
+	//       HashedNode fits into the picture, since Commit(), Commitment() and Hash()
+	//	     now panics. Despite these calls must not happen at runtime, it is still
+	//	     quite risky. The reason we end up in this place is because PBSS came quite
+	//	     recently compared with the VerkleNode interface design. We should probably
+	//	     reconsider splitting the interface or find some safer workaround.
 	panic("can not commit a hash node")
 }
 
--- a/hashednode.go
+++ b/hashednode.go
@@ -30,69 +30,48 @@ import (
 	"fmt"
 )
 
-type HashedNode struct {
-	commitment  []byte
-	cachedPoint *Point
-}
+type HashedNode struct{}
 
-func (*HashedNode) Insert([]byte, []byte, NodeResolverFn) error {
+func (HashedNode) Insert([]byte, []byte, NodeResolverFn) error {
 	return errInsertIntoHash
 }
 
-func (*HashedNode) Delete([]byte, NodeResolverFn) (bool, error) {
+func (HashedNode) Delete([]byte, NodeResolverFn) (bool, error) {
 	return false, errors.New("cant delete a hashed node in-place")
 }
 
-func (*HashedNode) Get([]byte, NodeResolverFn) ([]byte, error) {
+func (HashedNode) Get([]byte, NodeResolverFn) ([]byte, error) {
 	return nil, errors.New("can not read from a hash node")
 }
 
-func (n *HashedNode) Commit() *Point {
-	if n.commitment == nil {
-		panic("nil commitment")
-	}
-	if n.cachedPoint == nil {
-		n.cachedPoint = new(Point)
-		n.cachedPoint.SetBytesUncompressed(n.commitment, true)
-	}
-	return n.cachedPoint
+func (HashedNode) Commit() *Point {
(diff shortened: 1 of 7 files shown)


hashednode.go
```diff
@@ -30,69 +30,48 @@ import (
...
-		panic("nil commitment")
-	}
-	if n.cachedPoint == nil {
-		n.cachedPoint = new(Point)
-		n.cachedPoint.SetBytesUncompressed(n.commitment, true)
-	}
-	return n.cachedPoint
+func (HashedNode) Commit() *Point {
+	panic("can not commit a hash node")
 }
```
jsign, line 49: Here's a tricky situation I'm not comfortable with, but I was forced by the API signature.

There's no real way to return something meaningful and correct in this API. In the case of `Empty` we could at least return the identity element. Here that would be wrong.

The reality is that `HashedNode` is now very hand-wavy on being a `VerkleNode` in the current "shape and form" of `VerkleNode` interface asking for `Commit()`, `Commitment()` and `Hash()`.

In practice, these panics should never trigger since any node that is `HashedNode` will be resolved before calling these kinds of APIs; but in the theoretical world of what's possible, I see these panics as a big risk.

This might require some rethought of our `VerkleNode` interface (maybe splitting in two, etc). We could also return something else (that would be incorrect; like the identity element) just to avoid a panic. But quite honestly, if a call to these methods happens that's 100% a bug.

hashednode.go
```diff
@@ -30,69 +30,48 @@ import (
...
-	copy(c.commitment, n.commitment)
-	return c
+func (HashedNode) Copy() VerkleNode {
+	return HashedNode{}
 }
 
-func (n *HashedNode) toDot(parent, path string) string {
-	return fmt.Sprintf("hash%s [label=\"H: %x\"]\n%s -> hash%s\n", path, n.commitment, parent, path)
+func (HashedNode) toDot(parent, path string) string {
+	return fmt.Sprintf("hash%s \n%s -> hash%s\n", path, parent, path)
```
jsign, line 68: This API isn't a big deal since it's for debugging, but we can't print hashes here.

gballet (commented):
LGTM. Just waiting for your feedback on my `FlushedNode` comment, and I'll merge.

hashednode.go
jsign: I've never used this `toDot` to debug really, but I didn't know has been useful to you.

There's no way to do any resolving since `toDot` doesn't receive a `resolver` interface, so we can only print whatever we have in the parameters which is something I already do here. 

I added a `label="unresolved"` to at least signal that this node exists and is unresolved. If in the future we add a resolver to `toDot`, then we can resolve which is the commitment of this hashed node.

hashednode.go
jsign: Added.

gballet (approved):
LGTM

PR #377, opened 2023-07-11
Make GetProofItems capable of resolving hashed nodes

gballet:


tree.go
```diff
@@ -834,11 +847,14 @@ func (n *InternalNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]
 			// repeated as many times? It would be wasteful, so the
 			// decoding code has to be aware of this corner case.
 			esses = append(esses, extStatusAbsentEmpty|((n.depth+1)<<3))
-			pe.Vals = append(pe.Vals, nil)
+			for _ = range group {
+				// Append one nil value per key in this missing stem
+				pe.Vals = append(pe.Vals, nil)
+			}
```
jsign, line 853: Maybe this is known, but looks overlapping with the other PR.

tree.go
```diff
@@ -794,7 +794,20 @@ func (n *InternalNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]
 	for i, child := range n.children {
 		fiPtrs[i] = &fi[i]
 		if child != nil {
-			points[i] = child.Commitment()
+			var c VerkleNode
+			if _, ok := child.(HashedNode); ok {
```
jsign, line 798: Maybe add a comment TODO to remember to remove this when `BatchSerialize` doesn't hash nodes.

jsign (approved):

gballet:
I'm going to directly fix the open storage trie commit, it should be simpler than implementing this. This PR can still be needed in case the tree that is being proven can not fit in RAM, so I'm not closing it right away.

PR #379, opened 2023-07-18
Parallelize Commit() and converted leaf node insertions

jsign:
This PR parallelizes CPU (and some IO) heavy work since this makes a difference in machines, especially ones with slow CPU clocks (i.e: Rock5B).


conversion.go
```diff
@@ -95,6 +95,56 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
 }
 
 func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
+	sort.Slice(leaves, func(i, j int) bool {
+		return bytes.Compare(leaves[i].stem, leaves[j].stem) < 0
+	})
```
jsign, line 100: We first sort the _to be inserted leaves_ by stem, so we can group them by stem-first-byte by walking them serially.

conversion.go
```diff
@@ -95,6 +95,56 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
...
+				}
+				resolved, err := ParseNode(serialized, 1)
+				if err != nil {
+					return fmt.Errorf("parsing node %x: %w", serialized, err)
+				}
+				n.children[lastChildrenIdx] = resolved
+			}
+			n.cowChild(byte(lastChildrenIdx))
+		}
+	}
```
jsign, line 121: Here's the only subtle thing about this change. 
Some context first: the idea is that we'll insert all the leaves on each subtree for each children of the root node.

The underlying insertion in each subtree, at some point, will try to call `n.cowChild(X)` to mark the current children in the parent as "dirty" so future `Commit()` will do the proper work.

But, for the root node all the parallel calls will try to do `cowChild(...)` which isn't allowed since that is updating the cow-map, and concurrent read-write access to maps is a race condition in Go.

To avoid introducing locks, or any kind of `if n.depth == 0 { ...}` branches, we first do these `cowChild(...)` calls first before we parallelize the insertion work. So, we still let the future goroutines to call `cowChild(...)` but they'll find that children are already marked as dirty, so they will move on. That means that all `cowChild(...)` calls will be only read, thus there's no race problem.

This is a quite simple trick that is easy to see that is correctly implemented. But if there was a bug here, the geth will panic since Go automatically detects read-write concurrent access to maps and panics. Of course, that won't happen, but this kind of "potential race" isn't something that can go unnoticed at runtime.

conversion.go
```diff
@@ -95,6 +95,56 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
...
+	for i := range leaves {
+		if leaves[currStemFirstByte].stem[0] != leaves[i].stem[0] {
+			start := currStemFirstByte
+			end := i
+			group.Go(func() error {
+				return n.insertMigratedLeaves(leaves[start:end], resolver)
+			})
+			currStemFirstByte = i
+		}
+	}
```
jsign, line 136: We walk the leaves to be inserted waiting for the next leaf to have a different first-byte-stem. If that's the case, we take the current group and fire the insertion in a separate goroutine.

conversion.go
```diff
@@ -95,6 +95,56 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
...
+	group.Go(func() error {
+		return n.insertMigratedLeaves(leaves[currStemFirstByte:], resolver)
+	})
```
jsign, line 139: We fire the last group.

conversion.go
```diff
@@ -95,6 +95,56 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
...
+	if err := group.Wait(); err != nil {
+		return fmt.Errorf("inserting migrated leaves: %w", err)
+	}
```
jsign, line 142: We wait for all to finish, and we're done.

conversion.go
```diff
@@ -95,6 +95,56 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
...
+
+	return nil
+}
+
+func (n *InternalNode) insertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
```
jsign, line 147: This is a new private method that is basically doing the work that was done before. I extracted it to separate the parallelization meta-logic from the actual work. So we can see that the actual work is unaware of any kind of "avoid root node cowChild calls" or anything like that.

tree.go
```diff
@@ -683,62 +685,86 @@ func (n *InternalNode) Commit() *Point {
 	internalNodeLevels := make([][]*InternalNode, StemSize)
 	n.fillLevels(internalNodeLevels)
```
jsign, line 686: This diff looks more complicated than it should, I'd suggest looking in split mode.

tree.go
```diff
@@ -683,62 +685,86 @@ func (n *InternalNode) Commit() *Point {
...
-		// later.
-		for _, node := range nodes {
-			for idx, nodeChildComm := range node.cow {
-				points = append(points, nodeChildComm)
-				points = append(points, node.children[idx].Commitment())
-				cowIndexes = append(cowIndexes, int(idx))
+
+		minBatchSize := 4
+		if len(nodes) <= minBatchSize {
+			commitNodesAtLevel(nodes)
```
jsign, line 696: If the amount of nodes at this level is 4 or less, lets do the work directly to avoid firing goroutines and paying overhead.

tree.go
```diff
@@ -683,62 +685,86 @@ func (n *InternalNode) Commit() *Point {
...
+				start := i
+				end := i + batchSize
+				if end > len(nodes) {
+					end = len(nodes)
+				}
+				wg.Add(1)
+				go func() {
+					defer wg.Done()
+					commitNodesAtLevel(nodes[start:end])
+				}()
```
jsign, line 714: TL;DR: we divide the list of nodes to use at most all available cores, but each core at least doing minimum amount of work.

tree.go
```diff
@@ -683,62 +685,86 @@ func (n *InternalNode) Commit() *Point {
...
 		}
+	}
+	return n.commitment
+}
 
-		// We generate `frs` which will contain the result for each element in `points`.
-		frs := make([]*Fr, len(points))
-		for i := range frs {
-			frs[i] = &Fr{}
+func commitNodesAtLevel(nodes []*InternalNode) {
```
jsign, line 722: This method extracted the actual work done; it's identical to the previous code.

tree.go
```diff
@@ -683,62 +685,86 @@ func (n *InternalNode) Commit() *Point {
...
 
-		// We calculate the difference between each (new commitment - old commitment) pair, and store it
-		// in the same slice to avoid allocations.
-		for i := 0; i < len(frs); i += 2 {
-			frs[i/2].Sub(frs[i+1], frs[i])
-		}
-		// Now `frs` have half of the elements, and these are the Frs differences to update commitments.
-		frs = frs[:len(frs)/2]
+	// Do a single batch calculation for all the points in this level.
+	toFrMultiple(frs, points)
```
jsign, line 744: So this `toFrMultiple` before was a single call per level; now by the nature of the work-splitting we're doing N `toFrMultiple` calls. So that means some extra inversions in total, but using all cores which is much faster.

tree.go
```diff
@@ -683,62 +685,86 @@ func (n *InternalNode) Commit() *Point {
...
+	for _, node := range nodes {
+		poly := make([]Fr, NodeWidth)
+		for i := 0; i < len(node.cow); i++ {
+			poly[cowIndexes[cowIndex]] = *frs[frsIdx]
+			frsIdx++
+			cowIndex++
 		}
+		node.cow = nil
+		node.commitment.Add(node.commitment, cfg.CommitToPoly(poly, 0))
 	}
```
jsign, line 767: This is the heavy diff-updating part done by each goroutine.

tree.go
gballet: is there a known cut-off point after which it makes sense not to do too many `toFrMultiple` or is it already faster with 2 cores?

gballet (commented):
LGTM. Left a few questions but we can merge it today.

tree.go
jsign: 4 nodes means one inverse and a couple of microseconds per diff-commit updating, so _feels_ like a good magic number.

The reality is that this number probably depends on how fast the CPU clock is. The faster it is, the bigger `minBatchSize` might be to justify the overhead of spawning goroutines and sync them.

In the grand scheme of things, probably it isn't that important to tune it further since the bottleneck are layers that have thousands of nodes, so you'll be saturating all your cores anyway.

For example, at depth 1 most of the time we'll have to update 256 nodes since that's pretty obvious. In that case, `minBatchSize` doesn't matter much since unless you have 64 cores, your batch will have more than 4 elements. And the same applies to other layers that might have hundreds or thousands of nodes in the level, which is where most of the work is being done.

jsign: Another angle to explain this is: if we have 16 nodes to update, and 16 cores... I don't want to spin up 16 goroutines each doing 1 thing. That's too little work per core and a lot of goroutine-sync overhead.

A more reasonable strategy is using only 4 cores of the 16 available ones, and making those 4 ones do a reasonable amount of work. That's the gist of it.

gballet (approved):
LGTM

PR #383, opened 2023-08-09
tree: add new method GetAndLoadForProof

jsign:
This PR adds a new helper method that will assist clients in preloading the tree for proof generation.

See https://github.com/gballet/go-ethereum/pull/255

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -836,7 +836,7 @@ func groupKeys(keys keylist, depth byte) []keylist {
 }
 
 func (n *InternalNode) GetAndLoadForProof(key []byte, resolver NodeResolverFn) ([]byte, error) {
-	// Each internal node that is part of the proof needs to load all it's
+	// Each internal node that is part of the proof needs to load all its
 	// children since it's needed for proof openings.
 	childrenKey := make([]byte, n.depth+1)
 	copy(childrenKey, key[:n.depth])


tree.go
```diff
@@ -772,6 +772,35 @@ func groupKeys(keys keylist, depth byte) []keylist {
 	return groups
 }
 
+func (n *InternalNode) GetAndLoadForProof(key []byte, resolver NodeResolverFn) ([]byte, error) {
+	// Each internal node that is part of the proof needs to load all it's
```
gballet, line 776: its*

gballet (commented):
Left some comments. I am not sure I understand why you don't simply use `Get`?

jsign:
> Left some comments. I am not sure I understand why you don't simply use `Get`?

When you call `Get`, it will resolve all internal nodes _directly_ in the path to the corresponding leaf. As in, the minimal node loading needed to reach the desired leaf (as expected).

This method does something extra needed for proof generation. When we want to generate a proof for a leaf value, we also need all the internal node children in that path. That's because we need all the evaluations of the vector in that internal node to prove the opening in internal nodes. [This part of the PR code](https://github.com/gballet/go-verkle/pull/383/files#diff-050830ac4334170e8d335e1b95519052a3461c5b57a4ca38a0531e35a8f3d388R781-R792) is exactly doing this (also note some code comment a few lines above).

Doing all this at the Geth layer is very complicated and confusing. This API is providing a way of asking: "Load the tree for this key, and also all the auxiliary stuff that I'll need to generate a proof without any further resolving" (as you'd have with using `Get`).

That means that Geth can call this method for every key in the witness. After finishing with these simple calls it will _know_ the loaded tree is fully ready to generate a proof (without a resolver or anything extra for proof generation).

That's exactly the idea of the referenced Geth PR in the PR description. While the EVM is collecting accessed keys in the witness, we use this API to load in the background the needed tree with all the resolved stuff. When the EVM finishes, the Witness already has loaded all that it needs so Geth can use this tree to generate the proof.

tree.go
```diff
@@ -1339,8 +1368,8 @@ func (n *LeafNode) GetProofItems(keys keylist) (*ProofElements, []byte, [][]byte
 			if len(esses) == 0 {
 				esses = append(esses, extStatusAbsentOther|(n.depth<<3))
 				poass = append(poass, n.stem)
-				pe.Vals = append(pe.Vals, nil)
 			}
+			pe.Vals = append(pe.Vals, nil) // TODO(PR-temp): workaround proving bug.
```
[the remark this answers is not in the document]
jsign: https://github.com/gballet/go-verkle/pull/398

gballet:
ah right of course. LGTM.

gballet (approved):
LGTM

proof_ipa.go
```diff
@@ -60,14 +60,10 @@ func (vp *VerkleProof) Copy() *VerkleProof {
...
-	for i := range vp.OtherStems {
-		ret.OtherStems[i] = vp.OtherStems[i]
-	}
-
+	copy(ret.OtherStems, vp.OtherStems)
 	copy(ret.DepthExtensionPresent, vp.DepthExtensionPresent)
-	for i := range vp.CommitmentsByPath {
-		ret.CommitmentsByPath[i] = vp.CommitmentsByPath[i]
-	}
+	copy(ret.CommitmentsByPath, vp.CommitmentsByPath)
```
jsign, line 65: The new linter was complaining about this (unrelated to the PR), but it's still a good point and fixed.

PR #384, opened 2023-08-30
create proofs with post state

gballet:
Changes the signature of `MakeVerkleMultiProof` to accept an extra root: if that root isn't `nil`, it means that the post state of the tree is also added to the proof.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -58,12 +58,14 @@ type Proof struct {
 	Cs         []*Point        // commitments, sorted by their path in the tree
 	PoaStems   [][]byte        // stems proving another stem is absent
 	Keys       [][]byte
-	Values     [][]byte
+	PreValues  [][]byte
+	PostValues [][]byte
 }
 
 type SuffixStateDiff struct {
 	Suffix       byte      `json:"suffix"`
 	CurrentValue *[32]byte `json:"currentValue"`
+	NewValue     *[32]byte `json:"newValue"`
 }
 
 type SuffixStateDiffs []SuffixStateDiff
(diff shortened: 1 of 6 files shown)


gballet:
This is missing specific tests, btw

proof_ipa.go
```diff
@@ -79,36 +82,68 @@ func GetCommitmentsForMultiproof(root VerkleNode, keys [][]byte, resolver NodeRe
 	return root.GetProofItems(keylist(keys), resolver)
 }
 
-func MakeVerkleMultiProof(root VerkleNode, keys [][]byte, resolver NodeResolverFn) (*Proof, []*Point, []byte, []*Fr, error) {
+func getProofElementsFromTree(preroot, postroot VerkleNode, keys [][]byte, resolver NodeResolverFn) (*ProofElements, []byte, [][]byte, [][]byte, []*Point, [][]Fr, []*Fr, []byte, error) {
```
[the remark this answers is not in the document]
gballet: Note that when the post trie is `nil`, the old behavior is retained, and the extra fields basically return the same thing as what is found in the first return parameter.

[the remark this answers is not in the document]
gballet: So there's a need for API cleanup, but works well enough for now.

proof_ipa.go
```diff
@@ -79,36 +82,68 @@ func GetCommitmentsForMultiproof(root VerkleNode, keys [][]byte, resolver NodeRe
...
+		postvals  = make([][]byte, len(keys))
+	)
+	for i := range pe.Cis {
+		proof_cis = append(proof_cis, pe.Cis[i])
+		proof_fs = append(proof_fs, pe.Fis[i])
+		proof_ys = append(proof_ys, pe.Yis[i])
+		proof_zs = append(proof_zs, pe.Zis[i])
+	}
+
+	if postroot != nil {
```
gballet, line 114: This block concatenates the post- and pre-state data together, to be used in the proof.

proof_ipa.go
```diff
@@ -57,12 +58,14 @@ type Proof struct {
...
 	Keys       [][]byte
-	Values     [][]byte
+	PreValues  [][]byte
+	PostValues [][]byte
 }
 
 type SuffixStateDiff struct {
 	Suffix       byte      `json:"suffix"`
 	CurrentValue *[32]byte `json:"currentValue"`
+	NewValue     *[32]byte `json:"new_value"`
```
jsign, line 68: Open question: is there a reason we want a pointer to an array? Feels to me like an unnecessary level of indirection.

proof_ipa.go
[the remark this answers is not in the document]
jsign: Yeah, feels like an overloaded method... we can think of cleaning it up later.
Maybe add a `TODO` in this method with "clean up" or similar.

proof_ipa.go
```diff
@@ -79,36 +82,76 @@ func GetCommitmentsForMultiproof(root VerkleNode, keys [][]byte, resolver NodeRe
...
+			proof_fs = append(proof_fs, pe_post.Fis[i])
+			proof_ys = append(proof_ys, pe_post.Yis[i])
+			proof_zs = append(proof_zs, pe_post.Zis[i])
+		}
+
+		// Set the post values, if they are untouched, leave them `nil`
+		for i, v := range pe.Vals {
+			if !bytes.Equal(v, pe_post.Vals[i]) {
+				postvals[i] = v
+			}
```
jsign, line 136: [Checked](https://notes.ethereum.org/Si5tEWlMTYerkhG3fOIAMQ):
>     # Null means value not updated

proof_ipa.go
```diff
@@ -79,36 +82,76 @@ func GetCommitmentsForMultiproof(root VerkleNode, keys [][]byte, resolver NodeRe
...
-	// var vals [][]byte
-	// for _, k := range keys {
-	// 	// TODO at the moment, do not include the post-data
-	// 	// val, _ := root.Get(k, nil)
-	// 	// vals = append(vals, val)
-	// 	vals = append(vals, keyvals[string(k)])
-	// }
+	// 1..=3: proof elements of the pre-state trie for serialization,
+	// 4: values to be inserted in the post-state trie for serialization
+	// 5..=7: aggregated pre+post elements for proving
```
jsign, line 142: Nit: `5..=8`

Instant karma for non-idiomatic syntax :laughing:

proof_ipa.go
```diff
@@ -126,17 +169,30 @@ func MakeVerkleMultiProof(root VerkleNode, keys [][]byte, resolver NodeResolverF
...
 	}
 	return proof, pe.Cis, pe.Zis, pe.Yis, nil
 }
 
+// VerifyVerkleProofWithPreAndPostTrie takes a pre-state trie, a post-state trie and a list of keys, and verifies that
+// the provided proof verifies. If the post-state trie is `nil`, the behavior is the same as `VerifyVerkleProof`.
+func VerifyVerkleProofWithPreAndPostTrie(proof *Proof, preroot, postroot VerkleNode, keys [][]byte, resolver NodeResolverFn, tc *Config) bool {
+	_, _, _, _, proof_cis, _, proof_ys, proof_zs, err := getProofElementsFromTree(preroot, postroot, keys, resolver)
+	if err != nil {
+		return false
```
jsign, line 190: Uhm... maybe a TODO to change method signature to return an error.

proof_ipa.go
```diff
@@ -177,26 +233,34 @@ func SerializeProof(proof *Proof) (*VerkleProof, StateDiff, error) {
...
+			newsd.NewValue = (*[32]byte)(proof.PostValues[i])
 		default:
 			var aligned [32]byte
-			copy(aligned[:valueLen], proof.Values[i])
-			stemdiff.SuffixDiffs = append(stemdiff.SuffixDiffs, SuffixStateDiff{
-				Suffix:       key[31],
-				CurrentValue: (*[32]byte)(unsafe.Pointer(&aligned[0])),
-			})
+			copy(aligned[:valueLen], proof.PostValues[i])
+			newsd.NewValue = (*[32]byte)(unsafe.Pointer(&aligned[0]))
```
jsign, line 260: To be honest, I'd add a TODO here to remove the `unsafe` usage. 
If we do the change of removing the `*[32]byte` indirection, we can simply do a `copy` here or even directly using the copy one line above.

proof_ipa.go
```diff
@@ -177,26 +233,34 @@ func SerializeProof(proof *Proof) (*VerkleProof, StateDiff, error) {
...
 			copy(stemdiff.Stem[:], key[:31])
 		}
-		var valueLen = len(proof.Values[i])
+		stemdiff.SuffixDiffs = append(stemdiff.SuffixDiffs, SuffixStateDiff{Suffix: key[31]})
+		newsd := &stemdiff.SuffixDiffs[len(stemdiff.SuffixDiffs)-1]
+
+		var valueLen = len(proof.PreValues[i])
+		switch valueLen {
+		case 0:
+			// null value
```
jsign, line 242: Shouldn't deleted lines L183-185 be here?

proof_ipa.go
```diff
@@ -370,3 +442,36 @@ func TreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) {
...
+		var (
+			values     = make([][]byte, NodeWidth)
+			overwrites bool
+		)
+
+		for _, suffixdiff := range stemstatediff.SuffixDiffs {
+			if len(suffixdiff.NewValue) > 0 {
+				// If the prestate is present, it means that at least one
+				// post value will be non-nil. In this case, make a copy
+				// of the pre tree and update all post values.
```
jsign, line 461: I find this comment a bit confusing. Nothing about this `for` loop references the pre-state (i.e: `make a copy` or similar)

proof_test.go
```diff
@@ -20,7 +20,7 @@
 // OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
 // ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 // OTHER DEALINGS IN THE SOFTWARE.
-//
+//TestProofOfAbsenceStemVerify
```
jsign, line 23: Unintended?

jsign (approved):
LGTM, but I left some comments for your consideration since maybe some of them require code changes.

proof_ipa.go
gballet: the problem is that we want the encoding to be `null` if the value doesn't get changed. We could set a slice, as in https://go.dev/play/p/zTnpuXBpY3J, but it means that there could be non-32 bytes values.

gballet: actually it might be catcheable, but I won't do this in this PR. Created issue #388.

proof_ipa.go
gballet: ok, I'll use the zig syntax

proof_ipa.go
gballet: I think that I factored everything and so nothing should be in there, am I missing something?

jsign: nvm, it's on line L236. LGMT.

proof_ipa.go
jsign: Actually from Go 1.17 we should be able to do `(*[4]byte)(slice)`

proof_ipa.go
```diff
@@ -352,14 +429,14 @@ func TreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) {
...
+			if len(proof.PreValues[i]) == 0 {
 				// Skip the nil keys, they are here to prove
 				// an absence.
 				continue
 			}
 
 			if bytes.Equal(k[:31], info[string(p)].stem) {
-				values[k[31]] = proof.Values[i]
+				values[k[31]] = proof.PreValues[i]
 			}
```
jsign, line 440: I still feel this `if` is a bit strange. We're reconstructing the `values` to be assigned in the leaf when we'll soon create the LeafNode.

But see what we do in `CreatePath(...)`:
- We use that `value` slice [as expected](https://github.com/gballet/go-verkle/blob/master/tree.go#L442-L447).
- We do [some extra assignments from `stemInfo`](https://github.com/gballet/go-verkle/blob/master/tree.go#L462-L464).

Is the latter needed? Or a noop? Or maybe I'm missing something.

jsign (approved):
Left some comments for your consideration.

proof_ipa.go
```diff
@@ -79,36 +82,76 @@ func GetCommitmentsForMultiproof(root VerkleNode, keys [][]byte, resolver NodeRe
...
+func MakeVerkleMultiProof(preroot, postroot VerkleNode, keys [][]byte, resolver NodeResolverFn) (*Proof, []*Point, []byte, []*Fr, error) {
+	pe, es, poas, postvals, proof_cis, proof_fs, _, proof_zs, err := getProofElementsFromTree(preroot, postroot, keys, resolver)
+	if err != nil {
+		return nil, nil, nil, nil, fmt.Errorf("error getting proof elements: %w", err)
+	}
 
 	cfg := GetConfig()
-	mpArg := ipa.CreateMultiProof(tr, cfg.conf, pe.Cis, pe.Fis, pe.Zis)
+	tr := common.NewTranscript("vt")
+	mpArg := ipa.CreateMultiProof(tr, cfg.conf, proof_cis, proof_fs, proof_zs)
```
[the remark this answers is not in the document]
gballet: Sure, add it to the never-ending Todo list 😜

proof_ipa.go
```diff
@@ -79,36 +82,76 @@ func GetCommitmentsForMultiproof(root VerkleNode, keys [][]byte, resolver NodeRe
...
+		proof_cis = append(proof_cis, pe.Cis[i])
+		proof_fs = append(proof_fs, pe.Fis[i])
+		proof_ys = append(proof_ys, pe.Yis[i])
+		proof_zs = append(proof_zs, pe.Zis[i])
+	}
+
+	// if a post-state tree is present, merge its proof elements with
+	// those of the pre-state tree, so that they can be proved together.
+	if postroot != nil {
+		pe_post, _, _, err := GetCommitmentsForMultiproof(postroot, keys, resolver)
```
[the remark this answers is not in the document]
gballet: Well the opening themselves will not be integrated to the proof. But you are correct that they will be computed. This is reusing a function that does too much, and at the end values that are just read will find themselves in the proof. This is not *incorrect* but it's much more than necessary.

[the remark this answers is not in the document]
gballet: Not going to fix it in this PR, created #393 to track it.

proof_ipa.go
gballet: yeah I can't remember why that is, and removing it seems to work. I won't remove it in this PR but I'm making a note of it in #395

PR #389, opened 2023-09-05
mod: go-ipa panic removal

jsign:
This PR contains changes when updating to  (now merged) https://github.com/crate-crypto/go-ipa/pull/56.

TODO:
- [x] Switch to final mod reference when underlying PR is merged.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -28,6 +28,7 @@ package verkle
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"sort"
 	"unsafe"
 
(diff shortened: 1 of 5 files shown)


proof_ipa.go
```diff
@@ -137,7 +141,7 @@ func MakeVerkleMultiProof(root VerkleNode, keys [][]byte, resolver NodeResolverF
 	return proof, pe.Cis, pe.Zis, pe.Yis, nil
 }
 
-func VerifyVerkleProof(proof *Proof, Cs []*Point, indices []uint8, ys []*Fr, tc *Config) bool {
+func VerifyVerkleProof(proof *Proof, Cs []*Point, indices []uint8, ys []*Fr, tc *Config) (bool, error) {
```
jsign, line 144: A return of `false/true, nil` means that the proof verified or not verified cryptographically.
An error is more related to a formatting problem or some error not related to cryptography.

Distinguishing both things can be helpful to understand if the proof not passing might be because it is unsound, or it's because of some deserialization problem or incorrectly constructed.

@gballet, if you think that's not entirely useful and want to only return a `bool` or only a `error`; I'm OK with that and we can adjust a bit more here.

gballet (approved):
I have done something similar in my post proof PR but even though it will be a pain to rebase, I think it's worth putting it in a separate commit.

PR #390, opened 2023-09-08
proof: use safe deserialization method

jsign:
We should only use fast/unsafe serialization from trusted sources (i.e: db).

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -230,7 +230,7 @@ func DeserializeProof(vp *VerkleProof, statediff StateDiff) (*Proof, error) {
 	commitments = make([]*Point, len(vp.CommitmentsByPath))
 	for i, commitmentBytes := range vp.CommitmentsByPath {
 		var commitment Point
-		if err := commitment.SetBytesUnsafe(commitmentBytes[:]); err != nil {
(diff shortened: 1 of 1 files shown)


proof_ipa.go
```diff
@@ -282,7 +282,7 @@ type stemInfo struct {
 }
 
 // TreeFromProof builds a stateless tree from the proof
-func TreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) {
+func TreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // skipcq: GO-R1005
```
jsign, line 285: @gballet, [DeepSource thought it was a good idea today ](https://app.deepsource.com/gh/gballet/go-verkle/run/b42ab770-414f-49ec-a49c-48de24cae7f7/go/)to change their criteria and now was complaining about the cyclomatic complexity of this function since it consider is too complex. 

May be right, but right now you're around these zones and might not be a good idea to try to solve that in this PR (which is unrelated).

gballet: yeah saw that, thank you. Ideally we would fix it altogether but let's skip that for now.

gballet (approved):
LGTM

PR #396, opened 2023-09-12
Check for all returned errors

jsign:
This PR enables the `errcheck` linter, which doesn't allow to ignored returned errors.

Most code changes can be grouped into two buckets:
1. Tests fixes: nothing entirely interesting. I can agree that it can be annoying in some places to make the error checking since some APIs can only fail because of a `resolver`, which is not used in many places. But if these APIs fail for another reason in the future, it suddenly becomes relevant.
2. "Real code" fixes: maybe worth double-checking the code here, but there were many _interesting_ places where errors were ignored.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,6 +12,7 @@ run:
 linters:
   disable-all: true
   enable:
+    - errcheck
     - gofmt
     - staticcheck
     - deadcode
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,7 +15,6 @@ linters:
     - errcheck
     - gofmt
     - staticcheck
-    - deadcode
     - goconst
     - goimports
     - gosimple
@@ -23,7 +22,7 @@ linters:
     - ineffassign
     - misspell
     - unconvert
-    - varcheck
+    - unused
 
 linters-settings:
   gofmt:
--- a/benchs/main.go
+++ b/benchs/main.go
@@ -17,9 +17,9 @@ func main() {
 func benchmarkInsertInExisting() {
 	f, _ := os.Create("cpu.prof")
 	g, _ := os.Create("mem.prof")
-	pprof.StartCPUProfile(f)
+	_ = pprof.StartCPUProfile(f)
 	defer pprof.StopCPUProfile()
-	defer pprof.WriteHeapProfile(g)
+	defer func() { _ = pprof.WriteHeapProfile(g) }()
 	// Number of existing leaves in tree
 	n := 1000000
 	// Leaves to be inserted afterwards
@@ -34,7 +34,9 @@ func benchmarkInsertInExisting() {
 		// Generate set of keys once
 		for i := 0; i < total; i++ {
 			key := make([]byte, 32)
-			rand.Read(key)
+			if _, err := rand.Read(key); err != nil {
+				panic(err)
+			}
 			if i < n {
 				keys[i] = key
 			} else {
--- a/conversion.go
+++ b/conversion.go
@@ -147,7 +147,7 @@ func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeReso
 	return nil
 }
 
-func (n *InternalNode) insertMigratedLeavesSubtree(leaves []LeafNode, resolver NodeResolverFn) error {
+func (n *InternalNode) insertMigratedLeavesSubtree(leaves []LeafNode, resolver NodeResolverFn) error { // skipcq: GO-R1005
 	for i := range leaves {
 		ln := leaves[i]
 		parent := n
--- a/conversion.go
+++ b/conversion.go
@@ -55,7 +55,9 @@ func BatchNewLeafNode(nodesValues []BatchNewLeafNodeData) ([]LeafNode, error) {
 					c1c2frs[2*i], c1c2frs[2*i+1] = new(Fr), new(Fr)
 				}
 
-				banderwagon.BatchMapToScalarField(c1c2frs, c1c2points)
+				if err := banderwagon.BatchMapToScalarField(c1c2frs, c1c2points); err != nil {
+					return fmt.Errorf("mapping to scalar field: %s", err)
+				}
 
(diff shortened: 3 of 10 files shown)


.golangci.yml
```diff
@@ -12,17 +12,17 @@ run:
 linters:
   disable-all: true
   enable:
+    - errcheck
```
jsign, line 15: Enabling new linter.

.golangci.yml
```diff
@@ -12,17 +12,17 @@ run:
...
-    - deadcode
     - goconst
     - goimports
     - gosimple
     - govet
     - ineffassign
     - misspell
     - unconvert
-    - varcheck
+    - unused
```
jsign, line 25: Fixing this warning:
```
WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused. 
WARN [runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused.
```

benchs/main.go
```diff
@@ -34,7 +34,9 @@ func benchmarkInsertInExisting() {
 		// Generate set of keys once
 		for i := 0; i < total; i++ {
 			key := make([]byte, 32)
-			rand.Read(key)
+			if _, err := rand.Read(key); err != nil {
+				panic(err)
```
jsign, line 38: This is fine since it's benchmark code.

conversion.go
```diff
@@ -145,7 +147,7 @@ func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeReso
 	return nil
 }
 
-func (n *InternalNode) insertMigratedLeavesSubtree(leaves []LeafNode, resolver NodeResolverFn) error {
+func (n *InternalNode) insertMigratedLeavesSubtree(leaves []LeafNode, resolver NodeResolverFn) error { // skipcq: GO-R1005
```
jsign, line 150: I'll create an issue to resolve these DeepSource linter situation being reducing the cyclomatic complexity, or finding a way to ignore this linter check in general.

encoding.go
```diff
@@ -109,11 +109,17 @@ func parseLeafNode(serialized []byte, depth byte) (VerkleNode, error) {
...
 	ln.c2 = new(Point)
-	ln.c2.SetBytesUncompressed(serialized[leafC2CommitmentOffset:leafC2CommitmentOffset+banderwagon.UncompressedSize], true)
+	if err := ln.c2.SetBytesUncompressed(serialized[leafC2CommitmentOffset:leafC2CommitmentOffset+banderwagon.UncompressedSize], true); err != nil {
+		return nil, fmt.Errorf("setting c2 commitment: %w", err)
+	}
 	ln.commitment = new(Point)
-	ln.commitment.SetBytesUncompressed(serialized[leafCommitmentOffset:leafC1CommitmentOffset], true)
+	if err := ln.commitment.SetBytesUncompressed(serialized[leafCommitmentOffset:leafC1CommitmentOffset], true); err != nil {
+		return nil, fmt.Errorf("setting commitment: %w", err)
+	}
```
jsign, line 122: Very interesting case we didn't check errors.

proof_ipa.go
```diff
@@ -309,15 +309,21 @@ func DeserializeProof(vp *VerkleProof, statediff StateDiff) (*Proof, error) {
...
+		if err := multipoint.IPA.L[i].SetBytes(b[:]); err != nil {
+			return nil, fmt.Errorf("setting L[%d]: %w", i, err)
+		}
 	}
 	multipoint.IPA.R = make([]Point, IPA_PROOF_DEPTH)
 	for i, b := range vp.IPAProof.CR {
-		multipoint.IPA.R[i].SetBytes(b[:])
+		if err := multipoint.IPA.R[i].SetBytes(b[:]); err != nil {
+			return nil, fmt.Errorf("setting R[%d]: %w", i, err)
+		}
```
jsign, line 326: Ditto

tree.go
```diff
@@ -710,7 +715,11 @@ func (n *InternalNode) Commit() *Point {
 				wg.Add(1)
 				go func() {
 					defer wg.Done()
-					commitNodesAtLevel(nodes[start:end])
+					if err := commitNodesAtLevel(nodes[start:end]); err != nil {
+						// TODO: make Commit() return an error
+						panic(err)
+					}
```
jsign, line 721: @gballet, we have some tension here. `commitNodesAtLevel` error was ignored, but actually we should check it. But that would mean bubbling the error to `Commit()`.

For now, considering the work that we're doing for Kaustinen, we can panic here for a while. But at some point I think there won't be other option that bubbling the error.

There's already an issue to resolve all the panics in the repo, so this will be part of that work.

WDYT?

gballet (approved):
LGTM. Left a couple comments but I'm happy to merge right now.

benchs/main.go
```diff
@@ -17,9 +17,9 @@ func main() {
 func benchmarkInsertInExisting() {
 	f, _ := os.Create("cpu.prof")
 	g, _ := os.Create("mem.prof")
-	pprof.StartCPUProfile(f)
+	_ = pprof.StartCPUProfile(f)
 	defer pprof.StopCPUProfile()
-	defer pprof.WriteHeapProfile(g)
+	defer func() { _ = pprof.WriteHeapProfile(g) }()
```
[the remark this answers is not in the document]
jsign: Uhm, in which way? AFAIK, we never run this benchmark. Also this change is just to make the linter happy, it has no other side effects.

or maybe missing something!

tree.go
jsign: Not a big fan of %w unless we're sure we're OK with exposing the error as part of the API. But maybe that's the case.

benchs/main.go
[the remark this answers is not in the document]
gballet: as @jsign correctly points out, that was already there before.

PR #398, opened 2023-10-04
Proof generation panic bug

jsign:
This PR fixes a panic in proof generation in one of the border-cases of multiple keys landing in the same leaf node.

It also adds a test that reproduced the panic without the fix.

I found this while doing: https://github.com/gballet/go-verkle/pull/383


gballet (approved):
LGTM

PR #401, opened 2023-10-15
fix absent keys in proofs

jsign:
This PR proposes a fix to https://github.com/gballet/go-verkle/pull/400.

It fixes two things:
1. Verifier fix: rebuilding the tree from a proof that has an absent key (other) case.
2. Prover fix: leaf nodes shouldn't open C1/C2 for keys that don't match the leaf steam.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -183,6 +183,11 @@ type (
 		c1, c2     *Point
 
 		depth byte
+
+		// IsPOAStub indicates if this LeafNode is a proof of absence
+		// for a steam that isn't present in the tree. This flag is only
+		// true in the context of a stateless tree.
+		isPOAStub bool
 	}
 )
 
@@ -380,6 +385,10 @@ func (n *InternalNode) InsertStem(stem []byte, values [][]byte, resolver NodeRes
 		// splits.
 		return n.InsertStem(stem, values, resolver)
 	case *LeafNode:
+		if child.isPOAStub {
+			return errIsPOAStub
+		}
+
 		n.cowChild(nChild)
 		if equalPaths(child.stem, stem) {
 			return child.insertMultiple(stem, values)
@@ -444,11 +453,10 @@ func (n *InternalNode) CreatePath(path []byte, stemInfo stemInfo, comms []*Point
 				stem:       stemInfo.stem,
 				values:     nil,
 				depth:      n.depth + 1,
+				isPOAStub:  true,
 			}
 			n.children[path[0]] = newchild
 			comms = comms[1:]
-			newchild.c1 = new(Point)
-			newchild.c2 = new(Point)
 		case extStatusPresent:
 			// insert stem
 			newchild := &LeafNode{
@@ -527,6 +535,9 @@ func (n *InternalNode) GetStem(stem []byte, resolver NodeResolverFn) ([][]byte, 
 		// splits.
 		return n.GetStem(stem, resolver)
 	case *LeafNode:
+		if child.isPOAStub {
+			return nil, errIsPOAStub
+		}
 		if equalPaths(child.stem, stem) {
 			return child.values, nil
 		}
@@ -1024,6 +1035,10 @@ func (n *InternalNode) touchCoW(index byte) {
 }
 
 func (n *LeafNode) Insert(key []byte, value []byte, _ NodeResolverFn) error {
+	if n.isPOAStub {
+		return errIsPOAStub
+	}
+
 	if len(key) != StemSize+1 {
 		return fmt.Errorf("invalid key size: %d", len(key))
 	}
@@ -1278,6 +1293,10 @@ func (n *LeafNode) Delete(k []byte, _ NodeResolverFn) (bool, error) {
 }
 
 func (n *LeafNode) Get(k []byte, _ NodeResolverFn) ([]byte, error) {
+	if n.isPOAStub {
+		return nil, errIsPOAStub
+	}
+
 	if !equalPaths(k, n.stem) {
 		// If keys differ, return nil in order to
 		// signal that the key isn't present in the
@@ -1378,9 +1397,6 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
 	if err := StemFromBytes(&poly[1], n.stem); err != nil {
 		return nil, nil, nil, fmt.Errorf("error serializing stem '%x': %w", n.stem, err)
 	}
-	if err := banderwagon.BatchMapToScalarField([]*Fr{&poly[2], &poly[3]}, []*Point{n.c1, n.c2}); err != nil {
-		return nil, nil, nil, fmt.Errorf("batch mapping to scalar fields: %s", err)
-	}
 
 	// First pass: add top-level elements first
 	var hasC1, hasC2 bool
@@ -1396,6 +1412,22 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
 			}
 		}
 	}
+
+	// If this tree is a full tree (i.e: not a stateless tree), we know we have c1 and c2 values.
+	// Also, we _need_ them independently of hasC1 or hasC2 since the prover needs `Fis`.
+	if !n.isPOAStub {
+		if err := banderwagon.BatchMapToScalarField([]*Fr{&poly[2], &poly[3]}, []*Point{n.c1, n.c2}); err != nil {
+			return nil, nil, nil, fmt.Errorf("batch mapping to scalar fields: %s", err)
+		}
+	} else {
+		if hasC1 {
+			n.c1.MapToScalarField(&poly[2])
+		}
+		if hasC2 {
+			n.c2.MapToScalarField(&poly[3])
+		}
+	}
+
 	if hasC1 {
 		pe.Cis = append(pe.Cis, n.commitment)
 		pe.Zis = append(pe.Zis, 2)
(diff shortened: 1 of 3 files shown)


tree.go
```diff
@@ -439,6 +439,16 @@ func (n *InternalNode) CreatePath(path []byte, stemInfo stemInfo, comms []*Point
...
+			newchild := &LeafNode{
+				commitment: comms[0],
+				stem:       stemInfo.stem,
+				values:     nil,
+				depth:      n.depth + 1,
+			}
+			n.children[path[0]] = newchild
+			comms = comms[1:]
+			newchild.c1 = new(Point)
+			newchild.c2 = new(Point)
```
jsign, line 451: This mostly means "empty value", using the same behavior of `extStatusPresent`.

I'm not yet entirely convinced this is ideal, but maybe it is sufficient. Having a `new(Point)` is setting a value that a future "reader" will use as correct but this technically means "not known". But we don't have a `PartialLeafNode` type.

When I say "sufficient" I refer for a stateless client. This client will only need this LeafNode as a proof that a LeafNode with a different steam is here, instead of "something else" that is looking for. So, it _shouldn't_ ever really **need** to read C1/C2, which "is fine".

It's all quite subtle. I'm not sure it's worth adding another LeafNode type for that.

Note that this comment is unrelated with the fix; since the same comment applies with existing code in `extStatusPresent` (i.e: L466, etc below).

tree.go
```diff
@@ -1368,19 +1397,37 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
 	if err := StemFromBytes(&poly[1], n.stem); err != nil {
 		return nil, nil, nil, fmt.Errorf("error serializing stem '%x': %w", n.stem, err)
 	}
-	if err := banderwagon.BatchMapToScalarField([]*Fr{&poly[2], &poly[3]}, []*Point{n.c1, n.c2}); err != nil {
-		return nil, nil, nil, fmt.Errorf("batch mapping to scalar fields: %s", err)
-	}
```
jsign, line 1373: OK, so here's a vital correctness fix.

We can't **always** do this here. If we're calling this `GetProofItems(...)` in a stateless reconstructed tree, then `n.c1` and `n.c2` will be `nil`. (Remember that now we only set the `isPOAStub` marker; so we can't blindly assume the data is there).

tree.go
```diff
@@ -1368,19 +1397,37 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
...
+		// Note that keys might contain keys that don't correspond to this leaf node.
+		// We should only analize the inclusion of C1/C2 for keys corresponding to this
+		// leaf node stem.
+		if equalPaths(n.stem, key) {
+			hasC1 = hasC1 || (key[31] < 128)
+			hasC2 = hasC2 || (key[31] >= 128)
+			if hasC2 {
+				break
+			}
+		}
```
jsign, line 1416: Nothing new here, this is the original fix.

tree.go
```diff
@@ -380,10 +385,15 @@ func (n *InternalNode) InsertStem(stem []byte, values [][]byte, resolver NodeRes
...
-		n.cowChild(nChild)
 		if equalPaths(child.stem, stem) {
+			// We can't insert anything into a POA leaf node.
+			if child.isPOAStub {
+				return errIsPOAStub
+			}
+			n.cowChild(nChild)
 			return child.insertMultiple(stem, values)
 		}
+		n.cowChild(nChild)
```
jsign, line 396: Interestingly, `(n *InsertNode) Insert(...)` internally calls `InsertStem(...)` (and not `(*LeafNode) Insert(...)`, so we short-circuit here.

Note that unfortunately I had to put `n.cowChild(...)` in two places, to avoid an unwanted side effect of doing that and doing `return errIsPOAStub` later, which would be incorrect. (i.e: unwanted side-effect).

tree.go
```diff
@@ -518,6 +537,11 @@ func (n *InternalNode) GetStem(stem []byte, resolver NodeResolverFn) ([][]byte,
 		return n.GetStem(stem, resolver)
 	case *LeafNode:
 		if equalPaths(child.stem, stem) {
+			// We can't return the values since it's a POA leaf node, so we know nothing
+			// about its values.
+			if child.isPOAStub {
+				return nil, errIsPOAStub
+			}
```
jsign, line 544: Other important short-circuit. `(*InternalNode) Get(..)` internally calls `(*InternalNode) GetStem(...)` to get all the values and index later.

tree.go
```diff
@@ -1368,19 +1400,36 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
...
+		}
+	}
+
+	// If this tree is a full tree (i.e: not a stateless tree), we know we have c1 and c2 values.
+	// Also, we _need_ them independently of hasC1 or hasC2 since the prover needs `Fis`.
+	if !n.isPOAStub {
+		if err := banderwagon.BatchMapToScalarField([]*Fr{&poly[2], &poly[3]}, []*Point{n.c1, n.c2}); err != nil {
+			return nil, nil, nil, fmt.Errorf("batch mapping to scalar fields: %s", err)
 		}
```
jsign, line 1424: If this leaf node isn't a stub, we do the original thing that I removed from above. Nothing weird, since we are in a non-stateless tree.

tree.go
```diff
@@ -1368,19 +1400,36 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
...
+	} else if hasC1 || hasC2 || n.c1 != nil || n.c2 != nil {
+		// This LeafNode is a proof of absence stub. It must be true that
+		// both c1 and c2 are nil, and that hasC1 and hasC2 are false.
+		// Let's just check that to be sure, since the code below can't use
+		// poly[2] or poly[3].
+		return nil, nil, nil, fmt.Errorf("invalid proof of absence stub")
```
jsign, line 1430: If the `LeafNode` **is** a stub, let's be defensive and check some invariants:
- We can't conclude that `hasC1` is true. 
- Same with `hasC2`.
- Any `LeafNode` with `isPOAStub` was constructed with `nil` c1 and c2. So check that.

This is important since lines below L1433 and L1439 will try to use the `poly[2]` and `poly[3]` values. Which in this case aren't calculated (because we can't!). 

Maybe this check that I did is very paranoid; since indirectly those branches below have the "right" if cases. But my check is stronger: I don't care what the `hasC1` or `hasC2` evaluation was; those can never be ON in a stub. Period.

tree.go
```diff
@@ -380,10 +385,15 @@ func (n *InternalNode) InsertStem(stem []byte, values [][]byte, resolver NodeRes
 		// splits.
 		return n.InsertStem(stem, values, resolver)
 	case *LeafNode:
-		n.cowChild(nChild)
 		if equalPaths(child.stem, stem) {
+			// We can't insert anything into a POA leaf node.
+			if child.isPOAStub {
+				return errIsPOAStub
+			}
+			n.cowChild(nChild)
```
gballet, line 393: I don't quite like this comment, because it sounds like you can not insert in ANY type of PoA `LeafNode`, and the fact is that you can. I would rephrase as "you can't update leaf values of a PoA stem" or something like taht.

tree.go
gballet: right, in 402 I catch this even at the call site in `(*InternalNode)` so that I don't have to come here. What I don't see, is why you need to move this code instead of just using an `if` like you do anyway line 1421 ?

gballet: you can still check the invariant later, but apart from that I don't quite see what that changes :thinking: ?

gballet (commented):
LGTM overall, but left a few questions and comments

tree.go
jsign: > because it sounds like you can not insert in ANY type of PoA LeafNode, and the fact is that you can

You can't. In today's definition of `isPOAStub`, the only case is `true` is for a `LeafNode` that is only there to prove the absence of a stem. In this case, you can't insert any value.

Maybe you're thinking about potential future "stub" cases where leaf nodes are partially filled and you _might_ be able to insert things?

tree.go
jsign: My point is that we need to return before calling `n.cowChild(..)`.

If we do `n.cowChild(...)` (saying that this children as changed), and return an error later, it would be incorrect to mark that children as changed.

Not sure if that deserves a comment. It's just something that any API that fails should be aware of (i.e: assuming it will succeed doing stuff, and fail later without reverting the effects).

tree.go
jsign: > right, in 402 I catch this even at the call site in `(*InternalNode)` so that I don't have to come here.

Personally, I don't like that solution since if someone calls `(*LeafNode) GetProofItems(...)` that code would do the incorrect thing. You can say: "but usually, `GetProofItems(...)` gets call through an InternalNode". While that's true, there's nothing that prevents the caller to have some LeafNode and call this API. (As we do in many tests even). 

> What I don't see, is why you need to move this code instead of just using an `if` like you do anyway line 1421 ?

It's true that I didn't _need_ to do that, but I did so I can put in the same if the extra checks in the `else` case below. Also, putting this earlier wasn't needed since it's use later in the logic.

In any case, I can move it back and have two separate `ifs`.

tree.go
gballet: yes, the way you can is precisely when you are inserting a new leaf that has a common prefix with the PoA stem. So you can not insert _values_ into it, but you can insert a `LeafNode` into it (by producing intermediate `InternalNode`s).

jsign: Right, but my comment is inside `equalPaths(child.stem, stem)` so I'm not talking about that case, but inserting values.

jsign: In any case, I can add `insert values` in the comment if that helps.

jsign: Done.

PR #404, opened 2023-10-17
tree: getproofitems replace children with resolved node

jsign:
This PR makes the resolved nodes from `GetProofItems` replace the `HashedNodes`.

If we don't do this, this method can fail for unresolved nodes despite technically receiving a `resolver` via params. I'll point out the exact situation in the PR comment.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -873,6 +873,7 @@ func (n *InternalNode) GetProofItems(keys keylist, resolver NodeResolverFn) (*Pr
 				if err != nil {
 					return nil, nil, nil, err
 				}
+				n.children[i] = c
 			} else {
 				c = child
 			}


tree.go
```diff
@@ -873,6 +873,7 @@ func (n *InternalNode) GetProofItems(keys keylist, resolver NodeResolverFn) (*Pr
 				if err != nil {
 					return nil, nil, nil, err
 				}
+				n.children[i] = c
```
jsign, line 876: If we don't do this, the line below `pec, es, other, err := n.children[childIdx].GetProofItems(group, resolver)` (L930) would fail since `n.children[childrenIdx]` can be a `HashedNode` and `GetProofItems` isn't allowed.

gballet (approved):
Makes sense. Thanks for catching that. LGTM.

PR #405, opened 2023-10-18
proof: fix copying slice bug

jsign:
This PR fixes a bug in the proof serialization code.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -332,7 +332,8 @@ func DeserializeProof(vp *VerkleProof, statediff StateDiff) (*Proof, error) {
 
 	poaStems = make([][]byte, len(vp.OtherStems))
 	for i, poaStem := range vp.OtherStems {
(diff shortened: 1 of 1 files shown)


proof_ipa.go
```diff
@@ -332,7 +332,8 @@ func DeserializeProof(vp *VerkleProof, statediff StateDiff) (*Proof, error) {
 
 	poaStems = make([][]byte, len(vp.OtherStems))
 	for i, poaStem := range vp.OtherStems {
-		poaStems[i] = poaStem[:]
```
jsign, line 335: Doing `poaStem[:]` was incorrect. It's taking the `poaStem` **array** as a slice. But `poaStem` is always the same variable in the loop, so all `poaStems` items will be pointing to the same value.

The effect this had is that all `poaStems` returned have the same value.

gballet (approved):
yeah, that nonsense happens more than it should. Thanks for catching that! LGTM.

PR #406, opened 2023-10-20
proof: fix statediff copy bug

jsign:
The `Copy()` function missed copying the `newValue` field.

When merged, I'll create a PR in geth kaustinen branch to update `go-verkle`.


gballet (approved):
LGTM

PR #407, opened 2023-10-24
remove post-state openings in proof

jsign:
This PR changes the proof generation to avoid including post-state value openings in Multiproofs.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -138,50 +138,27 @@ func getProofElementsFromTree(preroot, postroot VerkleNode, keys [][]byte, resol
 		return nil, nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("error getting pre-state proof data: %w", err)
 	}
 
-	// List of points and vectors and indices used to generate the IPA proof
-	// If only a pre-state root is available, it's a simple copy, but if both
-	// pre- and post-state root are present, it's merging the two lists.
-	var (
-		proof_cis []*Point
-		proof_fs  [][]Fr
-		proof_zs  []uint8
-		proof_ys  []*Fr
-		postvals  = make([][]byte, len(keys))
-	)
-	for i := range pe.Cis {
-		proof_cis = append(proof_cis, pe.Cis[i])
-		proof_fs = append(proof_fs, pe.Fis[i])
-		proof_ys = append(proof_ys, pe.Yis[i])
-		proof_zs = append(proof_zs, pe.Zis[i])
-	}
-
 	// if a post-state tree is present, merge its proof elements with
 	// those of the pre-state tree, so that they can be proved together.
+	postvals := make([][]byte, len(keys))
 	if postroot != nil {
-		pe_post, _, _, err := GetCommitmentsForMultiproof(postroot, keys, resolver)
-		if err != nil {
-			return nil, nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("error getting post-state proof data: %w", err)
-		}
-
-		for i := range pe_post.Cis {
-			proof_cis = append(proof_cis, pe_post.Cis[i])
-			proof_fs = append(proof_fs, pe_post.Fis[i])
-			proof_ys = append(proof_ys, pe_post.Yis[i])
-			proof_zs = append(proof_zs, pe_post.Zis[i])
-		}
-
+		// keys were sorted already in the above GetcommitmentsForMultiproof.
 		// Set the post values, if they are untouched, leave them `nil`
-		for i, v := range pe.Vals {
-			if !bytes.Equal(v, pe_post.Vals[i]) {
-				postvals[i] = pe_post.Vals[i]
+		for i := range keys {
+			val, err := postroot.Get(keys[i], resolver)
+			if err != nil {
+				return nil, nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("error getting post-state value for key %x: %w", keys[i], err)
+			}
+			if !bytes.Equal(pe.Vals[i], val) {
+				postvals[i] = val
 			}
 		}
 	}
 
 	// [0:3]: proof elements of the pre-state trie for serialization,
 	// 3: values to be inserted in the post-state trie for serialization
-	// [4:8]: aggregated pre+post elements for proving
-	return pe, es, poas, postvals, proof_cis, proof_fs, proof_ys, proof_zs, nil
+	// [4:8]: aggregated elements for proving
+	return pe, es, poas, postvals, pe.Cis, pe.Fis, pe.Yis, pe.Zis, nil
 }
 
 func MakeVerkleMultiProof(preroot, postroot VerkleNode, keys [][]byte, resolver NodeResolverFn) (*Proof, []*Point, []byte, []*Fr, error) {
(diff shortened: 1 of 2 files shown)


proof_ipa.go
```diff
@@ -138,50 +138,27 @@ func getProofElementsFromTree(preroot, postroot VerkleNode, keys [][]byte, resol
...
-			if !bytes.Equal(v, pe_post.Vals[i]) {
-				postvals[i] = pe_post.Vals[i]
+		for i := range keys {
+			val, err := postroot.Get(keys[i], resolver)
+			if err != nil {
+				return nil, nil, nil, nil, nil, nil, nil, nil, fmt.Errorf("error getting post-state value for key %x: %w", keys[i], err)
+			}
+			if !bytes.Equal(pe.Vals[i], val) {
+				postvals[i] = val
 			}
```
jsign, line 154: This is something that can be optimizable if we allow the client to do the work.

The client (geth), after the EVM block is executed, has full knowledge of which are all the post-state values, so there's no need to do this L147-L153 here.

This current implementation isn't that bad, since actually `postroot` is already hot in memory (i.e: all nodes resolved) since Geth already updated the tree. So technically, it shouldn't need to resolve any node thus should be quite fast.

In any case, if that's convincing then maybe this is good enough and can avoid some coupling. But, would be worth double-checking.

proof_test.go
```diff
@@ -914,6 +914,33 @@ func TestProofVerificationWithPostState(t *testing.T) {
 
 			proof, _, _, _, _ := MakeVerkleMultiProof(root, postroot, data.keystoprove, nil)
 
+		keys:
```
jsign, line 917: I'm adding a bit more of checking here. I want to check that the generated pre-state and post-state values match what was originally created in the proofed trees.

proof_test.go
```diff
@@ -914,6 +914,33 @@ func TestProofVerificationWithPostState(t *testing.T) {
...
+			for i := range proof.Keys {
+				// Check that the pre-state value is the one that we originally inserted.
+				for j := range data.keys {
+					if bytes.Equal(proof.Keys[i], data.keys[j]) {
+						if !bytes.Equal(proof.PreValues[i], data.values[j]) {
+							t.Fatalf("pre-state value mismatch for key %x: %x != %x", data.keys[j], proof.PreValues[i], data.values[j])
+						}
+						break
+					}
+				}
```
jsign, line 927: Check that the pre-state value always matches with the original value that was used when created the tree for the test.

proof_test.go
```diff
@@ -914,6 +914,33 @@ func TestProofVerificationWithPostState(t *testing.T) {
...
+
+				for j := range data.updatekeys {
+					// The the key was updated then check that the post-state value is the updated value.
+					if bytes.Equal(proof.Keys[i], data.updatekeys[j]) {
+						if !bytes.Equal(proof.PostValues[i], data.updatevalues[j]) {
+							t.Fatalf("post-state value mismatch for key %x: %x != %x", data.updatekeys[j], proof.PostValues[i], data.updatevalues[j])
+						}
+						continue keys
+					}
+				}
```
jsign, line 937: For each key in the proof, look if this key was updated in the tree. If that's the case, the post-state value must match the corresponding updated value.

proof_test.go
```diff
@@ -914,6 +914,33 @@ func TestProofVerificationWithPostState(t *testing.T) {
...
+				// If the key was not updated then check that the post-state value is null.
+				if proof.PostValues[i] != nil {
+					t.Fatalf("post-state value mismatch for key %x: %x != nil", proof.Keys[i], proof.PostValues[i])
+				}
```
jsign, line 941: If we didn't find the `key` in the `updatedkeys`, this means we're proving a key that wasn't changed. This means that post-state must be `nil` (which semantically means that it is the same as the pre-state value).

proof_test.go
```diff
@@ -937,7 +964,12 @@ func TestProofVerificationWithPostState(t *testing.T) {
 			if err != nil {
 				t.Fatalf("error recreating post tree: %v", err)
 			}
-			if err = VerifyVerkleProofWithPreAndPostTrie(dproof, dpreroot, dpostroot); err != nil {
+			// Check that the reconstructed post-state tree root matches the real tree.
+			if !postroot.Commitment().Equal(dpostroot.Commitment()) {
+				t.Fatalf("differing root commitments %x != %x", dpostroot.Commitment().Bytes(), postroot.Commitment().Bytes())
+			}
```
jsign, line 970: An extra important check: check that the reconstructed post-state tree root ends up matching the original root.

proof_ipa.go
```diff
@@ -138,50 +138,27 @@ func getProofElementsFromTree(preroot, postroot VerkleNode, keys [][]byte, resol
...
-		proof_zs  []uint8
-		proof_ys  []*Fr
-		postvals  = make([][]byte, len(keys))
-	)
-	for i := range pe.Cis {
-		proof_cis = append(proof_cis, pe.Cis[i])
-		proof_fs = append(proof_fs, pe.Fis[i])
-		proof_ys = append(proof_ys, pe.Yis[i])
-		proof_zs = append(proof_zs, pe.Zis[i])
-	}
```
jsign, line 156: Now that we only use pre-state for Multiproofs, we don't need to do copy slices.

proof_ipa.go
gballet: yeah let's do it in the client, using the access witness if possible.

proof_ipa.go
```diff
@@ -138,50 +138,27 @@ func getProofElementsFromTree(preroot, postroot VerkleNode, keys [][]byte, resol
...
 			}
 		}
 	}
 
 	// [0:3]: proof elements of the pre-state trie for serialization,
 	// 3: values to be inserted in the post-state trie for serialization
-	// [4:8]: aggregated pre+post elements for proving
-	return pe, es, poas, postvals, proof_cis, proof_fs, proof_ys, proof_zs, nil
+	// [4:8]: aggregated elements for proving
+	return pe, es, poas, postvals, pe.Cis, pe.Fis, pe.Yis, pe.Zis, nil
```
gballet, line 161: you are already returning `pe` so no need to return `pe.Cis`, `pe.Fis`, `pe.Yis` and `pe.Zis`

```suggestion
	return pe, es, poas, postvals, nil
```

proof_ipa.go
```diff
@@ -226,10 +203,9 @@ func MakeVerkleMultiProof(preroot, postroot VerkleNode, keys [][]byte, resolver
 	return proof, pe.Cis, pe.Zis, pe.Yis, nil
 }
 
-// VerifyVerkleProofWithPreAndPostTrie takes a pre-state trie, a post-state trie and a list of keys, and verifies that
-// the provided proof verifies. If the post-state trie is `nil`, the behavior is the same as `VerifyVerkleProof`.
-func VerifyVerkleProofWithPreAndPostTrie(proof *Proof, preroot, postroot VerkleNode) error {
-	_, _, _, _, proof_cis, _, proof_ys, proof_zs, err := getProofElementsFromTree(preroot, postroot, proof.Keys, nil)
+// VerifyVerkleProofWithPreState takes a proof and a trusted tree root and verifies that the proof is valid.
+func VerifyVerkleProofWithPreState(proof *Proof, preroot VerkleNode) error {
+	_, _, _, _, proof_cis, _, proof_ys, proof_zs, err := getProofElementsFromTree(preroot, nil, proof.Keys, nil)
```
gballet, line 208: so yeah, get `pe` here and use it line 213. My suggested change above won't compile by itself.

gballet (commented):
so annoyed because it took me so much time to get it to work correctly :cry:

Anyhow, LGTM overall, just requested some changes in a function signature.

PR #410, opened 2023-10-26
proof: check poas slice usage exhaustion

jsign:
This PR adds an extra check when reconstructing the pre-state tree.

The tree reconstruction must exhaust using all the proof of absence stem list, if that isn't the case it means that the proof contains unused stems which must signal an invalid proof.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -449,6 +449,10 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 		}
 	}
 
+	if len(poas) != 0 {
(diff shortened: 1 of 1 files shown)


proof_ipa.go
```diff
@@ -449,6 +449,10 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
 		}
 	}
 
+	if len(poas) != 0 {
```
jsign, line 452: Considering in L419, `poas` is re-sliced on every usage this means that the final length must be zero.

gballet (approved):
LGTM

PR #412, opened 2023-11-01
Fix proof generation and verification code

jsign:
This PR:
- Adds more tests for proof generation and verification, mostly around border-cases.
- Adds more checks that should make proof be considered invalid.
- Merged #413
- Merged #414

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -407,6 +407,7 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 		poas  = proof.PoaStems
 	)
 
+	// The proof of absence stems must be sorted. If that isn't the case, the proof is invalid.
 	if !sort.IsSorted(bytesSlice(proof.PoaStems)) {
 		return nil, fmt.Errorf("proof of absence stems are not sorted")
 	}
@@ -421,24 +422,56 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 		}
 		switch si.stemType {
 		case extStatusAbsentEmpty:
+			// All keys that are part of a proof of absence, must contain empty
+			// pre and post state values. If that isn't the case, the proof is invalid.
+			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
+				if bytes.HasPrefix(k, path) {
+					if proof.PreValues[i] != nil {
+						return nil, fmt.Errorf("proof of absence (empty) stem %x has a value", si.stem)
+					}
+				}
+			}
 		case extStatusAbsentOther:
 			si.stem = poas[0]
 			poas = poas[1:]
+			// All keys that are part of a proof of absence, must contain empty
+			// pre and post state values. If that isn't the case, the proof is invalid.
+			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
+				if bytes.HasPrefix(k, si.stem) {
+					if proof.PreValues[i] != nil {
+						return nil, fmt.Errorf("proof of absence (other) stem %x has a value", si.stem)
+					}
+				}
+			}
 		default:
 			// the first stem could be missing (e.g. the second stem in the
 			// group is the one that is present. Compare each key to the first
 			// stem, along the length of the path only.
 			stemPath := stems[stemIndex][:len(path)]
 			si.values = map[byte][]byte{}
-			for i, k := range proof.Keys {
+			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
 				if bytes.Equal(k[:len(path)], stemPath) && proof.PreValues[i] != nil {
 					si.values[k[31]] = proof.PreValues[i]
 					si.has_c1 = si.has_c1 || (k[31] < 128)
 					si.has_c2 = si.has_c2 || (k[31] >= 128)
 					// This key has values, its stem is the one that
 					// is present.
-					si.stem = k[:31]
+					if si.stem == nil {
+						si.stem = k[:31]
+						continue
+					}
+					// Any other key with values must have the same
+					// same previously detected stem. If that isn't the case,
+					// the proof is invalid.
+					if !bytes.Equal(si.stem, k[:31]) {
+						return nil, fmt.Errorf("multiple keys with values found for stem %x", k[:31])
+					}
 				}
+			}
+			// For a proof of presence, we must always have detected a stem.
+			// If that isn't the case, the proof is invalid.
+			if si.stem == nil {
+				return nil, fmt.Errorf("no stem found for path %x", path)
 			}
 		}
 		info[string(path)] = si
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -423,7 +423,7 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 		switch si.stemType {
 		case extStatusAbsentEmpty:
 			// All keys that are part of a proof of absence, must contain empty
-			// pre and post state values. If that isn't the case, the proof is invalid.
+			// prestate values. If that isn't the case, the proof is invalid.
 			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
 				if bytes.HasPrefix(k, path) {
 					if proof.PreValues[i] != nil {
@@ -435,7 +435,7 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 			si.stem = poas[0]
 			poas = poas[1:]
 			// All keys that are part of a proof of absence, must contain empty
-			// pre and post state values. If that isn't the case, the proof is invalid.
+			// prestate values. If that isn't the case, the proof is invalid.
 			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
 				if bytes.HasPrefix(k, si.stem) {
 					if proof.PreValues[i] != nil {
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -404,6 +404,10 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 		poas  = proof.PoaStems
 	)
 
+	if !sort.IsSorted(bytesSlice(proof.PoaStems)) {
+		return nil, fmt.Errorf("proof of absence stems are not sorted")
+	}
+
 	// assign one or more stem to each stem info
 	for _, es := range proof.ExtStatus {
 		depth := es >> 3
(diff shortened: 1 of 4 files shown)


proof_ipa.go
```diff
@@ -387,7 +387,10 @@ type stemInfo struct {
 // PreStateTreeFromProof builds a stateless prestate tree from the proof.
 func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // skipcq: GO-R1005
 	if len(proof.Keys) != len(proof.PreValues) {
-		return nil, fmt.Errorf("incompatible number of keys and values: %d != %d", len(proof.Keys), len(proof.PreValues))
+		return nil, fmt.Errorf("incompatible number of keys and pre-values: %d != %d", len(proof.Keys), len(proof.PreValues))
+	}
+	if len(proof.Keys) != len(proof.PostValues) {
+		return nil, fmt.Errorf("incompatible number of keys and post-values: %d != %d", len(proof.Keys), len(proof.PostValues))
 	}
```
jsign, line 394: This should hold, which indirectly means `len(keys) == len(preStateKeys) == len(postStateKeys)`.

proof_ipa.go
```diff
@@ -404,6 +407,11 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
 		poas  = proof.PoaStems
 	)
 
+	// The proof of absence stems must be sorted. If that isn't the case, the proof is invalid.
+	if !sort.IsSorted(bytesSlice(proof.PoaStems)) {
+		return nil, fmt.Errorf("proof of absence stems are not sorted")
+	}
```
jsign, line 413: This is a check I pulled from the research repo. It's true that we must expect this slice to be sorted, so I do that here. I had to create a `bytesSlice` type (see bottom of this file).

proof_ipa.go
```diff
@@ -414,25 +422,57 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
 		case extStatusAbsentEmpty:
+			// All keys that are part of a proof of absence, must contain empty
+			// pre and post state values. If that isn't the case, the proof is invalid.
+			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
+				if bytes.HasPrefix(k, path) {
+					if proof.PreValues[i] != nil {
+						return nil, fmt.Errorf("proof of absence (empty) stem %x has a value", si.stem)
+					}
+				}
+			}
```
jsign, line 433: Explained in comments -- relevant check. Same with `extStateusAbsentOther` below.

proof_ipa.go
```diff
@@ -414,25 +422,57 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
+					if si.stem == nil {
+						si.stem = k[:31]
+						continue
+					}
+					// Any other key with values must have the same
+					// same previously detected stem. If that isn't the case,
+					// the proof is invalid.
+					if !bytes.Equal(si.stem, k[:31]) {
+						return nil, fmt.Errorf("multiple keys with values found for stem %x", k[:31])
+					}
```
jsign, line 468: OK, this was me being paranoid but it's an important check.

Before in our code, we always did `si.stem = k[:31]`. Technically, if there were multiple keys in the same stem, it will reassign to the same value. Totally fine since that's not very relevant.

Now, this extra check in L466 is relevant. If we find some key that apparently matches the stem-path that we're processing but the full stem is different, then something hairy is happening. 

Somebody is trying to inject a claim about a key that might go in this place in the tree, but doesn't correspond to this Leaf. In normal circumstances this shouldn't happen, since if that was the case, in this place in the tree we wouldn't have a LeafNode but an Internal node that would distinguish both paths. But if I manually craft a proof dishonestly, I can create this situation that should be rejected. (Note that if we're at this place, this injected key has a value non-nil in `Values`; so clearly something is really bad (i.e: this isn't a proof of absence case))

proof_ipa.go
```diff
@@ -414,25 +422,57 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
 				}
 			}
+			// For a proof of presence, we must always have detected a stem.
+			// If that isn't the case, the proof is invalid.
+			if si.stem == nil {
+				return nil, fmt.Errorf("no stem found for path %x", path)
+			}
```
jsign, line 475: Very important check. If we were in a proof of presence, we need to have detected the correct stem. If we end up with a stemInfo that doesn't have a `stem` value, then we'll create the `LeafNode` with an empty stem which is incorrect.

I believe the way we do this `si.stem` assignment is flawed; but I'll leave that convo for another PR when I try creating a test that will prove why that might be the case. For now, just see this check as a sanity check that clearly should be true (right?).

tree.go
```diff
@@ -448,6 +448,12 @@ func (n *InternalNode) CreatePath(path []byte, stemInfo stemInfo, comms []*Point
 			// unknown node.
 			n.children[path[0]] = Empty{}
 		case extStatusAbsentOther:
+			if len(comms) == 0 {
+				return comms, fmt.Errorf("missing commitment for stem %x", stemInfo.stem)
+			}
```
jsign, line 453: Added multiple `if len(comms) == 0` here and below. I'm just trying to prevent any panic situation when we do the `comm[1:]` reslicings.

tree.go
```diff
@@ -448,6 +448,12 @@ func (n *InternalNode) CreatePath(path []byte, stemInfo stemInfo, comms []*Point
...
+			if len(stemInfo.stem) != StemSize {
+				return comms, fmt.Errorf("invalid stem size %d", len(stemInfo.stem))
+			}
```
jsign, line 456: Again, sanity checks that we must always have a proper value here. If that isn't the case the `LeafNode { stem: ... }` value could be `nil` and that's wrong.

tree.go
```diff
@@ -483,6 +501,8 @@ func (n *InternalNode) CreatePath(path []byte, stemInfo stemInfo, comms []*Point
 			for b, value := range stemInfo.values {
 				newchild.values[b] = value
 			}
+		default:
+			return comms, fmt.Errorf("invalid stem type %d", stemInfo.stemType)
```
jsign, line 505: Probably will never happen since these cases depend on us and not external factors. But let's be safe.

proof_ipa.go
```diff
@@ -414,25 +422,57 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
 					si.has_c2 = si.has_c2 || (k[31] >= 128)
 					// This key has values, its stem is the one that
 					// is present.
-					si.stem = k[:31]
+					if si.stem == nil {
+						si.stem = k[:31]
+						continue
+					}
+					// Any other key with values must have the same
+					// same previously detected stem. If that isn't the case,
```
gballet, line 464:
```suggestion
					// Any other key with values must have the same
					// previously detected stem. If that isn't the case,
```

gballet (approved):
LGTM

tree.go
```diff
@@ -915,12 +935,16 @@ func (n *InternalNode) GetProofItems(keys keylist, resolver NodeResolverFn) (*Pr
...
+			addedStems := map[string]struct{}{}
 			for i := 0; i < len(group); i++ {
+				if _, ok := addedStems[string(group[i][:StemSize])]; !ok {
+					// A question arises here: what if this proof of absence
+					// corresponds to several stems? Should the ext status be
+					// repeated as many times? It's wasteful, so consider if the
+					// decoding code can be aware of this corner case.
+					esses = append(esses, extStatusAbsentEmpty|((n.depth+1)<<3))
+					addedStems[string(group[i][:StemSize])] = struct{}{}
+				}
```
jsign, line 947: I understand the point of the comment, but unfortunately it isn't that easy to do this without buying again all the corner cases.

I think the comment is still a fair take on the situation, so that's why I'm leaving it here, but if somebody wants to propose doing that optimization, it should:
- Do the optimization
- Add tests for that optimization
- Run all Kaustinen to triple-check stuff

Of course, the complicated point here is to asses if the amount of bytes we'd save in a real chain is meaningful for the complexity we'd be adding. (Since this is proposing breaking the 1:1 between stems and extension statuses).

tree.go
```diff
@@ -1492,21 +1521,22 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
...
+		// // at C1).
+		// if count == 0 {
+		// 	// TODO(gballet) maintain a count variable at LeafNode level
+		// 	// so that we know not to build the polynomials in this case,
+		// 	// as all the information is available before fillSuffixTreePoly
+		// 	// has to be called, save the count.
+		// 	esses = append(esses, extStatusAbsentEmpty|(n.depth<<3))
+		// 	pe.Vals = append(pe.Vals, nil)
+		// 	continue
+		// }
```
jsign, line 1539: I'm not sure this was the cause of issues, but definitely I think we need to chat about this.

This was adding a `extStatusAbsentEmpty`, and that's very tricky. 
I totally understand the intention that if we are proving **a single key** which lies in an empty half of the leaf node, then we could use `extStatusAbsentEmpty` since the verifier could assume there's no LeafNode here and get a `nil`.

Note that this is a trick and not representative of the real tree. The real tree **has** a LeafNode, but for this particular situation we can "simulate" that doesn't.

Let's assume for a moment we're OK with that.

Still, the code here isn't doing things correctly. We can add this `extStatusAbsentEmpty` if there's **a single key** proven in this way. As in, if for a key we did this addition of `extStatusAbsentEmpty`, and then the next key proves the presence in a non-empty half, then we should definitely removed that `extStatusAbsentEmpty` thing... if not, we'd have more than one extension status for keys of this leaf.

_Maybe_ it's something that we can try to optimize. But it's not correct to simply add this empty extension status. We need to have some internal state to track that we did that, and if anything else happens in a next key (i.e: proof of absence other, or presence) then we need to revert that addition.

We can keep chatting in Matrix about this. I just wanted to leave some comment since I have this fresh in my mind now.

tree.go
```diff
@@ -1533,9 +1563,12 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
 		pe.Yis = append(pe.Yis, &leaves[0], &leaves[1])
 		pe.Fis = append(pe.Fis, suffPoly[:], suffPoly[:])
 		pe.Vals = append(pe.Vals, n.values[key[31]])
-		if len(esses) == 0 || esses[len(esses)-1] != extStatusPresent|(n.depth<<3) {
+
+		if _, ok := addedStems[string(key[:StemSize])]; !ok {
 			esses = append(esses, extStatusPresent|(n.depth<<3))
+			addedStems[string(key[:StemSize])] = struct{}{}
 		}
```
jsign, line 1570: This previous (deleted) `if` was wrong. Now we always need to add the extension status for the present stem. (That `len(esses) == 0` is not correct, since we could have proof of absence extensions status now).

In short, we simply add once the extension status present.

tree.go
gballet: > Still, the code here isn't doing things correctly. We can add this extStatusAbsentEmpty if there's a single key proven in this way. As in, if for a key we did this addition of extStatusAbsentEmpty, and then the next key proves the presence in a non-empty half, then we should definitely removed that extStatusAbsentEmpty thing... if not, we'd have more than one extension status for keys of this leaf.

Yes, the extra empty was being removed if a proof of presence was found. That was line 1474, when setting `esses = nil`.

It's ok, let's keep things simple and have one extension status per stem.

tree.go
gballet: ok yeah that makes sense

PR #413, opened 2023-11-01
Proof of absence border case test

jsign:
This PR adds an extra tests case that proves, what I consider, an unfortunate flaw in an optimization included a while ago ~around #188.

The explanation is that if we have a proof of absence and a proof of presence in the same leaf node, **and** that proof of absence is for a `nil` value, we get in a situation in which we can't figure out the appropriate `stem`.

### DIFF
--- a/proof_test.go
+++ b/proof_test.go
@@ -1016,6 +1016,47 @@ func TestProofOfAbsenceBorderCase(t *testing.T) {
 	}
 }
 
+func TestProofOfAbsenceBorderCaseReversed(t *testing.T) {
+	root := New()
+
+	key1, _ := hex.DecodeString("0001000000000000000000000000000000000000000000000000000000000001")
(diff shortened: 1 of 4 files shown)


proof_test.go
```diff
@@ -1088,3 +1088,40 @@ func TestGenerateProofWithOnlyAbsentKeys(t *testing.T) {
 		}
 	}
 }
+
+func TestProofOfPresenceWithEmptyValue(t *testing.T) {
```
jsign, line 1092: This test case is pretty simple, and I believe shows a potentially fundamental flaw of the optimization done compared to the spec reference regarding "leaving out" the extension statuses for stems that have both proof of presence and absence.

The test case is simple:
- Add a key with a value.
- Prove that another key that corresponds to that `LeafNode` has value `nil`.

This fails for what looks like a simple to fix bug, but after you do the obvious fix it will be apparent that that will break another test and the fundamental issue will become more apparent. 

I can explain more in Matrix.

PR #414, opened 2023-11-02
Proof of absence and presence for the same path with nil proof of presence value

jsign:
This is a proposed fix to https://github.com/gballet/go-verkle/pull/413.

At the end of the day, it's almost aligned with the ultra-original Python implementation, which _probably_ did this for the same reason I found #413 problematic.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -412,12 +412,12 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 		return nil, fmt.Errorf("proof of absence stems are not sorted")
 	}
 
-	// We build a cache of stems that have a presence extension status.
-	stemsWithExtPresent := map[string]struct{}{}
+	// We build a cache of paths that have a presence extension status.
+	pathsWithExtPresent := map[string]struct{}{}
 	i := 0
 	for _, es := range proof.ExtStatus {
 		if es&3 == extStatusPresent {
-			stemsWithExtPresent[string(stems[i])] = struct{}{}
+			pathsWithExtPresent[string(stems[i][:es>>3])] = struct{}{}
 		}
 		i++
 	}
@@ -446,14 +446,14 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 			// If that is the case, we don't have to do anything since the corresponding leaf will be
 			// constructed by that extension status (already processed or to be processed).
 			// In other case, we should get the stem from the list of proof of absence stems.
-			if _, ok := stemsWithExtPresent[string(path)]; !ok {
+			if _, ok := pathsWithExtPresent[string(path)]; !ok {
 				si.stem = poas[0]
 				poas = poas[1:]
 			}
 			// All keys that are part of a proof of absence, must contain empty
 			// prestate values. If that isn't the case, the proof is invalid.
 			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
-				if bytes.HasPrefix(k, si.stem) {
+				if bytes.HasPrefix(k, stems[stemIndex]) {
 					if proof.PreValues[i] != nil {
 						return nil, fmt.Errorf("proof of absence (other) stem %x has a value", si.stem)
 					}
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -459,7 +459,7 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 					}
 				}
 			}
-		default:
+		case extStatusPresent:
 			si.values = map[byte][]byte{}
 			si.stem = stems[stemIndex]
 			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
@@ -474,6 +474,8 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 			if si.stem == nil {
 				return nil, fmt.Errorf("no stem found for path %x", path)
 			}
+		default:
+			return nil, fmt.Errorf("invalid extension status: %d", si.stemType)
 		}
 		info[string(path)] = si
 		paths = append(paths, path)
@@ -488,6 +490,9 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 				break
 			}
 		}
+	}
+	if stemIndex != len(stems) {
+		return nil, fmt.Errorf("not all stems were used: %d", len(stems))
 	}
 
 	if len(poas) != 0 {
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -401,8 +401,6 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 	if len(stems) != len(proof.ExtStatus) {
 		return nil, fmt.Errorf("invalid number of stems and extension statuses: %d != %d", len(stems), len(proof.ExtStatus))
 	}
-	// stemIndex := 0
-
 	var (
 		info  = map[string]stemInfo{}
 		paths [][]byte
@@ -482,21 +480,7 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 		}
 		info[string(path)] = si
 		paths = append(paths, path)
-
-		// // Skip over all the stems that share the same path
-		// // to the extension tree. This happens e.g. if two
-		// // stems have the same path, but one is a proof of
-		// // absence and the other one is present.
-		// stemIndex++
-		// for ; stemIndex < len(stems); stemIndex++ {
-		// 	if !bytes.Equal(stems[stemIndex][:depth], path) {
-		// 		break
-		// 	}
-		// }
 	}
-	// if stemIndex != len(stems) {
-	// 	return nil, fmt.Errorf("not all stems were used: %d", len(stems))
-	// }
 
 	if len(poas) != 0 {
 		return nil, fmt.Errorf("not all proof of absence stems were used: %d", len(poas))
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -398,7 +398,10 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { // 
 			stems = append(stems, k[:31])
 		}
 	}
-	stemIndex := 0
+	if len(stems) != len(proof.ExtStatus) {
+		return nil, fmt.Errorf("invalid number of stems and extension statuses: %d != %d", len(stems), len(proof.ExtStatus))
+	}
+	// stemIndex := 0
 
 	var (
 		info  = map[string]stemInfo{}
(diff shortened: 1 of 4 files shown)


proof_ipa.go
```diff
@@ -443,36 +459,23 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
 					}
 				}
 			}
-		default:
-			// the first stem could be missing (e.g. the second stem in the
-			// group is the one that is present. Compare each key to the first
-			// stem, along the length of the path only.
-			stemPath := stems[stemIndex][:len(path)]
+		case extStatusPresent:
```
jsign, line 462: Let's be safe, and add a proper `default` that fails.

proof_ipa.go
```diff
@@ -488,6 +491,9 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
 			}
 		}
 	}
+	if stemIndex != len(stems) {
+		return nil, fmt.Errorf("not all stems were used: %d", len(stems))
+	}
```
jsign, line 496: An even further sanity check that must be true!

proof_test.go
```diff
@@ -1125,3 +1125,56 @@ func TestProofOfPresenceWithEmptyValue(t *testing.T) {
 		t.Fatal("differing commitment for child #0")
 	}
 }
+
+func TestDoubleProofOfAbsence(t *testing.T) {
```
jsign, line 1129: I'm adding yet another case here.
The idea is to generate a proof of two keys (with different stems!) that end up in the same LeafNode.

proof_test.go
```diff
@@ -1125,3 +1125,56 @@ func TestProofOfPresenceWithEmptyValue(t *testing.T) {
...
+	if len(proof.PoaStems) != 1 {
+		t.Fatalf("invalid number of proof-of-absence stems: %d", len(proof.PoaStems))
+	}
+
+	// Despite we have two steams proved absent, we only need one extension status.
+	// A single extension status can already regenerate this branch, so the second stem
+	// can identify this branch was recreated.
+	if len(proof.ExtStatus) != 1 {
+		t.Fatalf("invalid number of proof-of-absence stems: %d", len(proof.PoaStems))
+	}
```
jsign, line 1179: Important checks. This is an optimization that doesn't exist in the spec, and is the thing I want to check with Dankrad. In the spec it adds two PoaStems and two ExtStatus.

tree.go
```diff
@@ -1488,10 +1488,11 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
 		// corner case (see previous corner case): if a proof-of-absence
 		// stem was found, and it now turns out the same stem is used as
 		// a proof of presence, clear the proof-of-absence list to avoid
-		// redundancy.
+		// redundancy. Note that we don't delete the extension statuses
+		// since that is needed to figure out which is the correct
+		// stem for this path.
 		if len(poass) > 0 {
 			poass = nil
-			esses = nil
```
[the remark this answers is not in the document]
gballet: yeah, that was added when I tried to only have one extension status to prove for two different stems (one present, one missing). It's ok to remove that.

gballet (approved):
LGTM. It's true that it makes things a bit simpler.

proof_test.go
jsign: Note that this change isn't for simplifying stuff. It's about correctness.

Saving that byte means you can't solve a border case. So it doesn't matter if we like it or not.

jsign: Saying it differently, if there's a way to save that byte and make it work I'm interested in knowing that.

jsign: Actually, it can be more than one byte if we have multiple stems proving absence in this same leaf node. One can argue what is the chance of this pattern happening multiple times, and my intuition says it's somewhat low.

In any case, at some point we have to come to a conclusion and if that's the case we need to change our impl to use more than one extension status, or change the spec.

proof_ipa.go
```diff
@@ -413,11 +413,11 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
 	}
 
 	// We build a cache of stems that have a presence extension status.
-	stemsWithExtPresent := map[string]struct{}{}
+	pathsWithExtPresent := map[string]struct{}{}
 	i := 0
 	for _, es := range proof.ExtStatus {
 		if es&3 == extStatusPresent {
-			stemsWithExtPresent[string(stems[i])] = struct{}{}
+			pathsWithExtPresent[string(stems[i][:es>>3])] = struct{}{}
```
jsign, line 420: This was a bug.

proof_test.go
```diff
@@ -1016,6 +1016,47 @@ func TestProofOfAbsenceBorderCase(t *testing.T) {
 	}
 }
 
+func TestProofOfAbsenceBorderCaseReversed(t *testing.T) {
+	root := New()
+
+	key1, _ := hex.DecodeString("0001000000000000000000000000000000000000000000000000000000000001")
+	key2, _ := hex.DecodeString("0000000000000000000000000000000000000000000000000000000000000001")
```
jsign, line 1023: Note this new test case which is exactly the same of `TestProofOfAbsenceBorderCase` but with key1 and key2 reversed. (Note: at some point we can refactor tests and make it sub-tests; for now I'm interested in correctness).

This fails because `key2` is the absent one, and is processed first than `key1` thus we add an extension status. But considering how tree reconstruction works, it will skip `key1`. (It's complicated to explain here; so I can explain better in Matrix).

proof_ipa.go
```diff
@@ -412,12 +412,12 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
 
-	// We build a cache of stems that have a presence extension status.
-	stemsWithExtPresent := map[string]struct{}{}
+	// We build a cache of paths that have a presence extension status.
+	pathsWithExtPresent := map[string]struct{}{}
 	i := 0
 	for _, es := range proof.ExtStatus {
 		if es&3 == extStatusPresent {
-			stemsWithExtPresent[string(stems[i])] = struct{}{}
+			pathsWithExtPresent[string(stems[i][:es>>3])] = struct{}{}
```
jsign, line 420: This was a bug. But even fixing that, I still think this for loop is wrong.

proof_ipa.go
```diff
@@ -398,8 +398,9 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
 			stems = append(stems, k[:31])
 		}
 	}
-	stemIndex := 0
```
jsign, line 401: We don't need this anymore, things got simplified.

proof_ipa.go
```diff
@@ -412,81 +413,73 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
+	pathsWithExtPresent := map[string]struct{}{}
+	i := 0
 	for _, es := range proof.ExtStatus {
-		depth := es >> 3
-		path := stems[stemIndex][:depth]
+		if es&3 == extStatusPresent {
+			pathsWithExtPresent[string(stems[i][:es>>3])] = struct{}{}
+		}
+		i++
+	}
```
jsign, line 424: Nothing new here compared to last review.

proof_ipa.go
```diff
@@ -412,81 +413,73 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
-			for i, k := range proof.Keys { // TODO: DoS risk, use map or binary search.
-				if bytes.HasPrefix(k, si.stem) {
-					if proof.PreValues[i] != nil {
-						return nil, fmt.Errorf("proof of absence (other) stem %x has a value", si.stem)
-					}
+			for j := range proof.Keys { // TODO: DoS risk, use map or binary search.
+				if bytes.HasPrefix(proof.Keys[j], stems[i]) && proof.PreValues[j] != nil {
+					return nil, fmt.Errorf("proof of absence (other) stem %x has a value", si.stem)
 				}
 			}
```
jsign, line 449: First, we check that all keys for this stem have `nil` values.
This must be the case since all are absent. If that's not true, wrong proof.

proof_ipa.go
```diff
@@ -412,81 +413,73 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
+			if _, ok := pathsWithExtPresent[string(path)]; ok {
+				continue
+			}
+
+			// Note that this path doesn't have proof of presence (previous if check above), but
+			// it can have multiple proof of absence. If a previous proof of absence had already
+			// created the stemInfo for this path, we don't have to do anything.
+			if _, ok := info[string(path)]; ok {
+				continue
+			}
```
jsign, line 464: Third, a small border case -- if a previous proof of absence already created this path, we don't have to do anything.

proof_ipa.go
```diff
@@ -412,81 +413,73 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
+
+			si.stem = poas[0]
+			poas = poas[1:]
```
jsign, line 467: Forth, we're the first (or only) proof of absence (without proof of presences) for this path. Create it.

proof_ipa.go
```diff
@@ -412,81 +413,73 @@ func PreStateTreeFromProof(proof *Proof, rootC *Point) (VerkleNode, error) { //
...
-					if !bytes.Equal(si.stem, k[:31]) {
-						return nil, fmt.Errorf("multiple keys with values found for stem %x", k[:31])
-					}
 				}
 			}
-			// For a proof of presence, we must always have detected a stem.
-			// If that isn't the case, the proof is invalid.
-			if si.stem == nil {
-				return nil, fmt.Errorf("no stem found for path %x", path)
-			}
```
jsign, line 475: All this got quite simplified since:
- si.stem is `stem[i]` directly due to the current 1:1 correspondence
- Look for keys, and simply add them into stemInfo (we don't care if their values are nil or not-nil. just add the values and mark c1/c2 appropriately).

tree.go
```diff
@@ -1464,34 +1464,39 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
...
-			// The list of extension statuses has to be of
-			// length 1 at this level, so skip otherwise.
+			// If this is the first extension status added for this path,
+			// add the proof of absence stem (only once). If later we detect a proof of
+			// presence, we'll clear the list since that proof of presence
+			// will be enough to provide the stem.
 			if len(esses) == 0 {
-				esses = append(esses, extStatusAbsentOther|(n.depth<<3))
 				poass = append(poass, n.stem)
 			}
```
jsign, line 1481: What comment says.

tree.go
```diff
@@ -1464,34 +1464,39 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
...
+			// Add an extension status absent other for this stem.
+			// Note we keep a cache to avoid adding the same stem twice (or more) if
+			// there're multiple keys with the same stem.
+			if _, ok := addedAbsentStems[string(key[:StemSize])]; !ok {
+				esses = append(esses, extStatusAbsentOther|(n.depth<<3))
+				addedAbsentStems[string(key[:StemSize])] = struct{}{}
+			}
```
jsign, line 1488: We use the `addedAbsentStems` to avoid adding duplicate extension statuses for the same stem. This can happen since this external loop is iterating `keys`, and not "grouped stems".

gballet (approved):
Thanks for tackling this, it was long overdue. Before we merge, have you ensured that it could sync kaustinen and verify every proof?

jsign:
> Thanks for tackling this, it was long overdue. Before we merge, have you ensured that it could sync kaustinen and verify every proof?

I'll merge all back to #412, and I'll test Kaustinen with that branch.

PR #415, opened 2023-11-03
Jsign proof tests

gballet:
same as #413 but against master so that CI runs.


jsign:
Closing since I'm merging stuff all into #412.

PR #417, opened 2023-11-17
fix: typo of NewLeafNodeWithNoComms()

agnxsh:
Fixed the typo in https://github.com/ethereum/go-verkle/blob/master/tree.go#L312

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -309,7 +309,7 @@ func NewLeafNode(stem []byte, values [][]byte) (*LeafNode, error) {
 	}, nil
 }
 
-// NewLeafNodeWithNoComms create a leaf node but does compute its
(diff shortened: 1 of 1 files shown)


jsign:
@agnxsh, I think this PR was correct? 

Actually, this makes me realize this is an API that I'd like not to be public -- feels to brittle. Anyway, that's another topic.

agnxsh:
> @agnxsh, I think this PR was correct?
> 
> Actually, this makes me realize this is an API that I'd like not to be public -- feels to brittle. Anyway, that's another topic.

oh sorry for closing it, i thought it was a trivial change, just caught my eye while going through the codebase, i can make a pr again

agnxsh:
https://github.com/ethereum/go-verkle/pull/420 here's the revised pr

PR #419, opened 2023-11-23
test: add random tree operation tests

weiihann:
This PR adds a random tree operation test case, referring [Geth's TestRandom implementation](https://github.com/ethereum/go-ethereum/blob/master/trie/trie_test.go#L616-L623). Minor modifications are as follows:
- only 32 bytes of random keys are created
- omitting `opItercheckhash` and `opNodeDiff`, can add them once the functionalities are ready

### DIFF
--- a/tree_test.go
+++ b/tree_test.go
@@ -32,12 +32,17 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 	mRand "math/rand"
+	"reflect"
 	"runtime"
 	"sort"
 	"strings"
 	"testing"
+	"testing/quick"
 	"time"
+
+	"github.com/davecgh/go-spew/spew"
 )
 
 // a 32 byte value, as expected in the tree structure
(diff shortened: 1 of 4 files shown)


tree_test.go
```diff
@@ -1649,3 +1654,127 @@ func TestLeafNodeInsert(t *testing.T) {
...
+	opInsert = iota
+	opDelete
+	opGet
+	opHash
+	opCommit
+	opProve
+	opMax
+)
+
+func (randTest) Generate(r *mRand.Rand, size int) reflect.Value {
```
jsign, line 1677: Do you mind adding a comment explaining that this `Generate(...)` method is defined since it's used indirectly by `testing/quick`. Maybe people not used to this package might infer this `Generate(...)` method is never used.

tree_test.go
```diff
@@ -1649,3 +1654,127 @@ func TestLeafNodeInsert(t *testing.T) {
...
+	}
+	return reflect.ValueOf(generateSteps(finishedFn, r))
+}
+
+func generateSteps(finished func() bool, r io.Reader) randTest {
+	var allKeys [][]byte
+	var one = make([]byte, 32)
+	genKey := func() []byte {
+		r.Read(one)
+		if len(allKeys) < 2 || one[0]%100 > 90 {
```
jsign, line 1690: Maybe add a comment to help the reader, like: ~"If these are the first two operations, we force inserting keys. If that isn't the case, make add a new key 10% of the time or return an existing key otherwise".

tree_test.go
```diff
@@ -1649,3 +1654,127 @@ func TestLeafNodeInsert(t *testing.T) {
...
+	var finishedFn = func() bool {
+		size--
+		return size == 0
+	}
+	return reflect.ValueOf(generateSteps(finishedFn, r))
+}
+
+func generateSteps(finished func() bool, r io.Reader) randTest {
+	var allKeys [][]byte
+	var one = make([]byte, 32)
```
jsign, line 1687: - Can we use a better name than `one`? Like `tmp`?
- Do we need a slice of 32 bytes? As far as I see, we always use one byte. But maybe I'm missing something.

tree_test.go
```diff
@@ -1649,3 +1654,127 @@ func TestLeafNodeInsert(t *testing.T) {
...
+			r.Read(key)
+			allKeys = append(allKeys, key)
+			return key
+		}
+		// use existing key
+		idx := int(one[0]) % len(allKeys)
+		return allKeys[idx]
+	}
+	var steps randTest
+	for !finished() {
```
jsign, line 1702: Quick question just in case: in `Generate(...)` the `finished` func definition first does `size--` and then `return size == 0`. 

This means that if `Generate(...)` is called with `size == 1`, then the `for` loop in this line (L1702) will return `false`, even if hasn't iterated once.

Not saying it is wrong, but feels more like a behavior for `size == 0` than `size == 1`. Just asking of this might not be wanted.

tree_test.go
```diff
@@ -1649,3 +1654,127 @@ func TestLeafNodeInsert(t *testing.T) {
...
+		r.Read(one)
+		step := randTestStep{op: int(one[0]) % opMax}
```
jsign, line 1704: Nit: Maybe rename `opMax` to `numOps`? At first I thought that `opMax` was another operation (since it's name was `op*`), but I see here you used it as a ~`len(numOperations)`.

No strong opinions, just a comment here.

tree_test.go
```diff
@@ -1649,3 +1654,127 @@ func TestLeafNodeInsert(t *testing.T) {
...
+		switch step.op {
+		case opInsert:
+			step.key = genKey()
+			step.value = make([]byte, 8)
+			binary.BigEndian.PutUint64(step.value, uint64(len(steps)))
```
jsign, line 1709: Curious about the logic of using the len of steps, instead of something like `r.Read(step.value)`? And also why `step.value` has length `8` instead of `32`?

Not even sure that changing to what I said would be better really -- just asking if there was a reason that I might be missing on why length is 8 and value is `len(steps)`.

tree_test.go
```diff
@@ -1649,3 +1654,127 @@ func TestLeafNodeInsert(t *testing.T) {
...
+			}
+			root.Commit()
+			proof, cis, zis, yis, _ := MakeVerkleMultiProof(root, nil, keys, nil)
+			if ok, err := VerifyVerkleProof(proof, cis, zis, yis, cfg); !ok || err != nil {
+				rt[i].err = fmt.Errorf("could not verify verkle proof: %s, err %v", ToDot(root), err)
+			}
+		case opHash:
+			root.Hash()
+		case opCommit:
+			root.Commit()
```
jsign, line 1761: Maybe for these two cases we can check that the returned pointer isn't `nil` since that should never happen.

Makes me feel we should actually not return pointers really. But that prob will have some effects in go-ethereum that we we'd prefer not to handle it now -- maybe we can also add here a `// TODO: reconsider if we should avoid returning pointers in Hash() and Commit()`.

jsign (commented):
Thank you!

Left some comments for your consideration.

If you can please take a look at the CI lints, there're a couple of things to fix that are quite easy. You can run the linter locally by running `go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.54.2 run` if helps.

After the changes I can re-review again to approve.

tree_test.go
weiihann: I don't think there are any particular reasons that it must be a length of `8` (aside from the fact that we are using `PutUint64`) and use the `len(steps)` as the value, since what it does is just trying to create some arbitrary values.

But I do think that it's better to just create random values by using `r.Read(step.value)`.

jsign (approved):

PR #421, opened 2023-12-02
refactor(tree): use const and type declaration

weiihann:
This PR does the following:
1. Use `Stem []byte` instead of `[]byte`, as well as `KeyToStem(key []byte)` instead of `key[:31]` in most occurrences that require stem. This enhances code readability, without affecting the underlying logic.
2. Replaces `32` with constants `LeafValueSize` or `KeySize` for code maintainability.

**BREAKING CHANGE**:
The `GetProofItems` function in the `VerkleNode` interface now returns `[]Stem` instead of `[][]byte`, which affects geth's usage of this library

Benchmark results:
```
                                        old allocs/op   new allocs/op  delta               
CommitLeaves/insert/leaves/1000-8         16.82k ± ∞ ¹   16.96k ± ∞ ¹  +0.87% (p=0.008 n=5)
CommitLeaves/insert/leaves/10000-8        155.4k ± ∞ ¹   155.3k ± ∞ ¹  -0.11% (p=0.008 n=5)
CommitFullNode-8                          3.795k ± ∞ ¹   3.795k ± ∞ ¹       ~ (p=1.000 n=5) ²
ModifyLeaves-8                            164.4k ± ∞ ¹   164.3k ± ∞ ¹       ~ (p=1.000 n=5)
```

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -13,7 +13,7 @@ import (
 
 // BatchNewLeafNodeData is a struct that contains the data needed to create a new leaf node.
 type BatchNewLeafNodeData struct {
-	Stem   []byte
+	Stem   Stem
 	Values map[byte][]byte
 }
 
--- a/debug.go
+++ b/debug.go
@@ -36,7 +36,7 @@ type (
 	}
 
 	ExportableLeafNode struct {
-		Stem   []byte   `json:"stem"`
+		Stem   Stem     `json:"stem"`
 		Values [][]byte `json:"values"`
 
 		C  [32]byte `json:"commitment"`
--- a/tree.go
+++ b/tree.go
@@ -55,6 +55,15 @@ func (kl keylist) Swap(i, j int) {
 	kl[i], kl[j] = kl[j], kl[i]
 }
 
+type Stem []byte
+
+func KeyToStem(key []byte) Stem {
+	if len(key) < StemSize {
+		panic(fmt.Errorf("key length (%d) is shorter than expected stem size (%d)", len(key), StemSize))
+	}
+	return Stem(key[:StemSize])
+}
+
 type VerkleNode interface {
 	// Insert or Update value into the tree
 	Insert([]byte, []byte, NodeResolverFn) error
@@ -80,7 +89,7 @@ type VerkleNode interface {
 	// returns them breadth-first. On top of that, it returns
 	// one "extension status" per stem, and an alternate stem
 	// if the key is missing but another stem has been found.
-	GetProofItems(keylist, NodeResolverFn) (*ProofElements, []byte, [][]byte, error)
+	GetProofItems(keylist, NodeResolverFn) (*ProofElements, []byte, []Stem, error)
 
 	// Serialize encodes the node to RLP.
 	Serialize() ([]byte, error)
@@ -176,7 +185,7 @@ type (
 	}
 
 	LeafNode struct {
-		stem   []byte
+		stem   Stem
 		values [][]byte
 
 		commitment *Point
@@ -254,7 +263,7 @@ func NewStatelessInternal(depth byte, comm *Point) VerkleNode {
 }
 
 // New creates a new leaf node
-func NewLeafNode(stem []byte, values [][]byte) (*LeafNode, error) {
+func NewLeafNode(stem Stem, values [][]byte) (*LeafNode, error) {
 	cfg := GetConfig()
 
 	// C1.
@@ -312,7 +321,7 @@ func NewLeafNode(stem []byte, values [][]byte) (*LeafNode, error) {
 // NewLeafNodeWithNoComms create a leaf node but does not compute its
 // commitments. The created node's commitments are intended to be
 // initialized with `SetTrustedBytes` in a deserialization context.
-func NewLeafNodeWithNoComms(stem []byte, values [][]byte) *LeafNode {
+func NewLeafNodeWithNoComms(stem Stem, values [][]byte) *LeafNode {
 	return &LeafNode{
 		// depth will be 0, but the commitment calculation
 		// does not need it, and so it won't be free.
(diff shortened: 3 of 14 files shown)


jsign (commented):
Thank you for the PR!

You did a good job describing the benefits of the change, and we've considered doing something like this before. Unfortunately, it comes with an unfortunate important side-effect which is more allocations, and impacts performance.

In the current (`master`) version, since we use `[]byte` this is a light reference to an existing place in memory, so we don't have to do these `copy(...)` in many places. These copies end up in arrays that escape to the heap, and thus count as heap allocations and adds pressure to the GC.

In the past, we had to be very strict with avoiding allocations to improve the performance of this library, so I'm not sure we can do any change that makes a performance regression on that front without being sure we'll need to revert this change later in the future.

To be a bit more objective, here's a benchmark difference between `master` and this branch (showing only relevant lines with perf change):
```
name                                      old allocs/op  new allocs/op  delta
CommitLeaves/insert/leaves/1000-16           17.0k ± 0%     18.5k ± 0%   +8.91%  (p=0.016 n=4+5)
CommitLeaves/insert/leaves/10000-16           155k ± 0%      176k ± 0%  +12.93%  (p=0.008 n=5+5)
CommitFullNode-16                            3.80k ± 0%     4.06k ± 0%   +6.74%  (p=0.008 n=5+5)
ModifyLeaves-16                               162k ± 0%      192k ± 0%  +18.45%  (p=0.008 n=5+5)
```
This isn't surprising by the nature of the change, since as I expected, this would mean more allocations and thus more GC work.

I think one option is using a feature since Go 1.17 to make `ExtStem` be a `*[31]byte`, and point that to the key. (More about this in the [spec](https://go.dev/ref/spec) in the section _Conversions from slice to array or array pointer_). So basically, it's an array of 31 bytes that will point to the same location in memory, so avoiding the performance problem. But I'm not sure using pointers to arrays just to signal a length of 31 is worth it.

Another option that still avoids the allocation, and gets most of the type system checking and semantics, is to define `type ExtStem { stem []byte }`. Despite this doesn't have the type-length internally, you could have a `KeyToStem(key []byte) ExtStem`, which would be mostly a `ExtStem { stem: key[:31] }` so "by definition", if we always use `KeyToStem(...)` (and nobody plays tricks creating `ExtStem` directly), we're much safer. And every place that needs to have a stem value would use `ExtStem` which already signals readers what this type means, it's length (checked at creation time), and this still avoids "copying bytes". 

We could also use a more relaxed `type ExtStem []byte`, but in function calls for `func something(a []ExtStem)` you could actually pass any `[]byte` as a parameter. Still isn't that bad since the user is seeing that it should pass `ExtStem` but it allows coercion at the language level.

cc @gballet if also has some preference around this refactor in general.

conversion.go
```diff
@@ -13,7 +12,7 @@ import (
 
 // BatchNewLeafNodeData is a struct that contains the data needed to create a new leaf node.
 type BatchNewLeafNodeData struct {
-	Stem   []byte
+	Stem   ExtStem
```
gballet, line 15: Is there a reason not to simply call that `Stem` ?

debug.go
```diff
@@ -36,7 +36,7 @@ type (
 	}
 
 	ExportableLeafNode struct {
-		Stem   []byte   `json:"stem"`
+		Stem   ExtStem  `json:"stem"`
```
gballet, line 39: not sure if that works out of the box (it might, haven't tested it). Just to make sure, please add a test that deserializes this structure in `debug_test.go`.

gballet (commented):
I get the argument for increased readability, but before we even consider merging the newer version of this PR (given its huge footprint), we need to be a 100% sure that there are no speed penalty for that. 

So if you want to proceed with that `ExtStem{ ... }` approach, make sure you have benchmarks to back it up.

tree.go
```diff
@@ -55,6 +55,15 @@ func (kl keylist) Swap(i, j int) {
 	kl[i], kl[j] = kl[j], kl[i]
 }
 
+type Stem []byte
+
+func KeyToStem(key []byte) Stem {
+	if len(key) < StemSize {
+		panic(fmt.Errorf("key length (%d) is shorter than expected stem size (%d)", len(key), StemSize))
+	}
```
jsign, line 63: I'm not sure if this is entirely useful, because if this `if` didn't exist we'd get a panic anyway in the next line. Maybe I'd accept it if this returns an error (that shouldn't happen, but at least avoid panicking the process).

@gballet, what's your opinion?

jsign (approved):
It's a bit unfortunate that `type Stem []byte` can't achieve what your original intention was, since we can literally put anything there as before. But, at least we have some extra semantical "signaling" out there...

The constant additions/changes were good. :+1: 

Left approved, but I'd like @gballet final take on this.

Thank you again!

tree.go
gballet: sorry I missed this question that seems to be blocking. On the one hand, panicking in production is extremely bad, so an error would be preferred, but at the moment a panic is more useful because it will indicate where this happens. If a stem that is smaller than 31 bytes is used, this is a huge bug that should never hit production. So I'm partial to leaving it as is for the moment.

tree.go
```diff
@@ -55,6 +55,15 @@ func (kl keylist) Swap(i, j int) {
 	kl[i], kl[j] = kl[j], kl[i]
 }
 
+type Stem []byte
+
+func KeyToStem(key []byte) Stem {
+	if len(key) < StemSize {
+		panic(fmt.Errorf("key length (%d) is shorter than expected stem size (%d)", len(key), StemSize))
```
gballet, line 62:
```suggestion
		panic(fmt.Errorf("key length (%d) is shorter than the expected stem size (%d)", len(key), StemSize))
```

gballet (approved):
LGTM. Thanks.

tree.go
```diff
@@ -1070,15 +1080,17 @@ func (n *LeafNode) Insert(key []byte, value []byte, _ NodeResolverFn) error {
 	if len(key) != StemSize+1 {
 		return fmt.Errorf("invalid key size: %d", len(key))
 	}
-	if !bytes.Equal(key[:StemSize], n.stem) {
-		return fmt.Errorf("stems doesn't match: %x != %x", key[:StemSize], n.stem)
+
+	stem := KeyToStem(key)
+	if !bytes.Equal(stem, n.stem) {
+		return fmt.Errorf("stems doesn't match: %x != %x", stem, n.stem)
```
gballet, line 1086:
```suggestion
		return fmt.Errorf("stems don't match: %x != %x", stem, n.stem)
```

PR #425, opened 2024-01-17
Fix Delete(...) early return

jsign:
This PR fixes a bug that appeared once in a while in the `TestRandom(...)` tests.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -618,7 +618,7 @@ func (n *InternalNode) Delete(key []byte, resolver NodeResolverFn) (bool, error)
 			// as well.
 			for _, c := range n.children {
 				if _, ok := c.(Empty); !ok {
-					break
+					return false, nil
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


tree.go
```diff
@@ -618,7 +618,7 @@ func (n *InternalNode) Delete(key []byte, resolver NodeResolverFn) (bool, error)
 			// as well.
 			for _, c := range n.children {
 				if _, ok := c.(Empty); !ok {
-					break
+					return false, nil
```
jsign, line 621: While breaking was reasonable, the right thing is to return `false`. Only `break` would mean L625 being executed signaling that a node deletion should happen, when that shouldn't be the case.

tree_test.go
```diff
@@ -1806,3 +1806,36 @@ func TestRandom(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestRandomExtracted(t *testing.T) {
```
jsign, line 1810: So after a while of debugging `TestRandom(...)` I extracted (and manually "minified") a simple test that surfaced the bug. Leaving the test here to avoid regressions.

Interestingly, this bug wasn't affecting [this other test](https://github.com/ethereum/go-verkle/blob/master/tree_test.go#L290-L358) that was targeting this API.

gballet (approved):
That was subtle. I guess it's a good thing we added this test. LGTM.

PR #426, opened 2024-01-19
mod: rename module to new location

jsign:
Rename module to use the new location. After this PR is merged, I'll fix the go-ethereum side of mod importing.


gballet (approved):
LGTM

PR #428, opened 2024-02-17
chore: Add `HashPointToBytes ` method

kevaundray:
## Rationale

We currently have two ways to convert a Point to 32 bytes, using the `.Bytes` method and the `MapToScalarField` method. The latter requires you to serialize the scalar field, though you can think of it as 32 bytes.  

~~For all intents and purposes in the verkle API, it seems we only need one of these. The root node uses the former, but for other internal nodes we use the latter. This PR changes it so that the root node is not treated differently from the other internal nodes and this allows us to remove `.Bytes` from the public API*. ~~

It is still used when serializing proofs, but from the point of view of a verkle trie library, a proof is(should be) just an opaque set of bytes.

*This PR alone won't allow us to remove `.Bytes` -- it is used in pedersen_hash/get_tree_key_hash. I'll make a PR to remove it from geth.

Linking to https://github.com/crate-crypto/rust-verkle/issues/86

EDIT: 

Ignacio pointed out that Serialize/Bytes is needed because we need to deserialize the root node in the header in order to use it in a proof.

### DIFF
--- a/ipa.go
+++ b/ipa.go
@@ -60,3 +60,9 @@ func FromBytes(fr *Fr, data []byte) {
 	copy(aligned[32-len(data):], data)
 	fr.SetBytes(aligned[:])
 }
+
+func MapToScalarFieldBytes(point *Point) [32]byte {
+	var hashedPoint Fr
+	point.MapToScalarField(&hashedPoint)
+	return hashedPoint.BytesLE()
+}
--- a/ipa.go
+++ b/ipa.go
@@ -61,7 +61,7 @@ func FromBytes(fr *Fr, data []byte) {
 	fr.SetBytes(aligned[:])
 }
 
-func MapToScalarFieldBytes(point *Point) [32]byte {
+func HashPointToBytes(point *Point) [32]byte {
 	var hashedPoint Fr
 	point.MapToScalarField(&hashedPoint)
 	return hashedPoint.BytesLE()
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -5,4 +5,4 @@ name = "go"
 enabled = true
 
   [analyzers.meta]
-  import_root = "github.com/gballet/go-verkle"
+  import_root = "github.com/ethereum/go-verkle"
--- a/.github/workflows/block_replay.yml
+++ b/.github/workflows/block_replay.yml
@@ -27,7 +27,7 @@ jobs:
       run: |
         cd geth
         cp /home/devops/verkle-test-dependencies/fork.txt .
-        go mod edit -replace=github.com/gballet/go-verkle=$PWD/..
+        go mod edit -replace=github.com/ethereum/go-verkle=$PWD/..
         go mod edit -replace=github.com/crate-crypto/go-ipa=github.com/crate-crypto/go-ipa@`grep go-ipa ../go.mod | cut -d'-' -f 5`
         go mod download github.com/crate-crypto/go-ipa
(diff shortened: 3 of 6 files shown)


tree.go
```diff
@@ -1618,7 +1618,10 @@ func (n *LeafNode) Value(i int) []byte {
 func (n *LeafNode) toDot(parent, path string) string {
 	var hash Fr
 	n.Commitment().MapToScalarField(&hash)
-	ret := fmt.Sprintf("leaf%s [label=\"L: %x\nC: %x\nStem: %x\nC₁: %x\nC₂:%x\"]\n%s -> leaf%s\n", path, hash.Bytes(), n.commitment.Bytes(), n.stem, n.c1.Bytes(), n.c2.Bytes(), parent, path)
+	// TODO: hash.Bytes() and MapToScalarFieldBytes(n.commitment) should give the same value
+	// TODO: its not clear why both values were needed before this change
+	// TODO: hash.Bytes should likely use BytesLE to be consistent with LE format
```
kevaundray, line 1623: nit: remove TODO

ipa.go
```diff
@@ -60,3 +60,9 @@ func FromBytes(fr *Fr, data []byte) {
 	copy(aligned[32-len(data):], data)
 	fr.SetBytes(aligned[:])
 }
+
+func MapToScalarFieldBytes(point *Point) [32]byte {
```
kevaundray, line 64: Whether it should go into go-ipa can be bikeshedded imo -- main goal is to update all libraries to have this breaking change

tree_test.go
```diff
@@ -1075,7 +1075,7 @@ var (
 		{197, 210, 70, 1, 134, 247, 35, 60, 146, 126, 125, 178, 220, 199, 3, 192, 229, 0, 182, 83, 202, 130, 39, 59, 123, 250, 216, 4, 93, 133, 164, 112},
 		{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
 	}
-	testAccountRootCommRust, _ = hex.DecodeString("10ed89d89047bb168baa4e69b8607e260049e928ddbcb2fdd23ea0f4182b1f8a")
+	testAccountRootCommRust, _ = hex.DecodeString("4caf7631af042c1615845a0116e789de0beea3e07f4b947377101aa822964614")
```
kevaundray, line 1078: This has been changed in this PR: https://github.com/crate-crypto/rust-verkle/pull/85

ipa.go
kevaundray, line 64: Bikeshedding: HashPointToBytes/PointToHashBytes

kevaundray: Would also add the LE postfix

kevaundray:
The `MapToScalarFIeldBytes` method is now only added so that it can be used downstream for get_tree_key_hash/pedersen_hash

ipa.go
jsign: Considering we return `[32]byte` and not `Fr`, then sounds more like `HashPointToBytes` or similar is a better name than referring to "ScalarField" which seems sounds leaking details. Also agree on LE.

jsign (approved):
LGTM. Left my 2c reg the API name.

jsign (approved):

PR #431, opened 2024-04-25
use special encoding for EoA

gballet:


gballet:
clearly something went wrong with the rebase, starting over

PR #432, opened 2024-04-25
special serialization for EoAs

gballet:
modifies the (de)serialization code in order to recognize an EoA and pack it more efficiently than a standard group.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -50,12 +50,14 @@ const (
 	internalCommitmentOffset = internalBitlistOffset + bitlistSize
 
 	// Leaf node offsets.
-	leafSteamOffset        = nodeTypeOffset + nodeTypeSize
-	leafBitlistOffset      = leafSteamOffset + StemSize
+	leafStemOffset         = nodeTypeOffset + nodeTypeSize
+	leafBitlistOffset      = leafStemOffset + StemSize
 	leafCommitmentOffset   = leafBitlistOffset + bitlistSize
 	leafC1CommitmentOffset = leafCommitmentOffset + banderwagon.UncompressedSize
 	leafC2CommitmentOffset = leafC1CommitmentOffset + banderwagon.UncompressedSize
 	leafChildrenOffset     = leafC2CommitmentOffset + banderwagon.UncompressedSize
+	leafBalanceSize        = 32
+	leafNonceSize          = 8
 )
 
 func bit(bitlist []byte, nr int) bool {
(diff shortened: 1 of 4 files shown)


encoding.go
```diff
@@ -78,10 +80,12 @@ func ParseNode(serializedNode []byte, depth byte) (VerkleNode, error) {
 	}
 
 	switch serializedNode[0] {
-	case leafRLPType:
```
jsign, line 81: Nit: should we add an extra line after L75 explaining EoAccount encoding?

encoding.go
```diff
@@ -123,6 +127,27 @@ func parseLeafNode(serialized []byte, depth byte) (VerkleNode, error) {
...
+	values[2] = nonce[:] // nonce
+	values[3] = emptyCodeHash[:]
+	values[4] = zero32[:] // 0 code size
+	ln := NewLeafNodeWithNoComms(serialized[leafStemOffset:leafStemOffset+StemSize], values[:])
+	ln.setDepth(depth)
+	ln.c1 = new(Point)
+	ln.c1.SetBytesUncompressed(serialized[leafStemOffset+StemSize:leafStemOffset+StemSize+banderwagon.UncompressedSize], true)
+	ln.c2 = &banderwagon.Identity
+	ln.commitment = new(Point)
+	ln.commitment.SetBytesUncompressed(serialized[leafStemOffset+StemSize+banderwagon.UncompressedSize:leafStemOffset+StemSize+banderwagon.UncompressedSize*2], true)
```
jsign, line 147: Nit: would be good to define constants for these `leafEoAC1Offset` and `leafEoACommOffset` to be sure same constants are used for serialization?

tree.go
```diff
@@ -1772,17 +1783,55 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
 				children = append(children, padding...)
 			}
 		}
+
+		if isEoA {
```
jsign, line 1787: I see, so in theory this should be named `maybeIsEoA` since someone could craft some storage slots to match this pattern. But should be fine.

(this is a jk comment, `isEoA` is fine)

tree.go
```diff
@@ -1772,17 +1783,55 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
...
+			switch i {
+			case 0:
+				isEoA = v != nil && bytes.Equal(v, zero32[:])
+			case 1:
+				// Balance should not be nil
+				isEoA = v != nil
+			case 2:
+				// last 24 bytes should be 0
+				isEoA = v != nil && bytes.Equal(v[8:32], zero24[:])
```
jsign, line 1796:
```suggestion
				isEoA = v != nil && bytes.Equal(v[leafNonceSize:], zero24[:])
```

tree.go
```diff
@@ -1772,17 +1783,55 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
...
+			case 3:
+				// code hash should be the empty code hash, but let's be tolerant
+				// and accept 0 as well.
+				isEoA = v != nil && (bytes.Equal(v, emptyCodeHash[:]) || bytes.Equal(v, zero32[:]))
```
jsign, line 1800: Not sure we can be tolerant with zero here, since in the parsing method we hard code an `emptyCodeHash` value for `v[3]`.

gballet: yeah I am reverting this to find out if this is what caused the issue I'm looking into.

PR #433, opened 2024-04-26
optimize serialization of single-slot leaves

gballet:
Another serialization optimization, for the case of a single leaf.

This could be extended to a group of leaves.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -145,17 +145,13 @@ func parseEoAccountNode(serialized []byte, depth byte) (VerkleNode, error) {
 	ln := NewLeafNodeWithNoComms(serialized[leafStemOffset:leafStemOffset+StemSize], values[:])
 	ln.setDepth(depth)
 	ln.c1 = new(Point)
-	err := ln.c1.SetBytesUncompressed(serialized[leafStemOffset+StemSize:leafStemOffset+StemSize+banderwagon.UncompressedSize], true)
-	if err != nil {
-		panic(err)
-		// return nil, err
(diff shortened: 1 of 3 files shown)


encoding.go
```diff
@@ -86,6 +88,8 @@ func ParseNode(serializedNode []byte, depth byte) (VerkleNode, error) {
 		return CreateInternalNode(serializedNode[internalBitlistOffset:internalCommitmentOffset], serializedNode[internalCommitmentOffset:], depth)
 	case eoAccountType:
 		return parseEoAccountNode(serializedNode, depth)
+	case singleSlotType:
```
jsign, line 91: Add bullet to method's comment. (Be aware that base branch has a similar comment)

tree.go
```diff
@@ -1811,8 +1815,15 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
 	// Create the serialization.
 	var result []byte
 	switch {
-	// TODO case for a group containing a single slot
-	// case count == 1:
+	case count == 1:
+		baseSize := nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafValueIndexSize + leafSlotSize
+		result = make([]byte, baseSize)
```
jsign, line 1820: Similar suggestion as in base branch. `baseSize` is constant, so we can avoid `make` and prob avoid an allocation by using an array.

jsign (approved):
LGTM, left some comments.

PR #435, opened 2024-05-29
make BatchSerialize not return an error

jwasinger:
Afaict, it should be safe to assume that a tree being serialized is not malformed.  Otherwise, it is catastrophic (programming error) and we might as well let it panic.


gballet (commented):
That sounds like a bad idea in general. Sure finding this means that something is critically wrong, but returning the error would increase the avenues we have for debugging. What is the reason why you want to panic instead?

jwasinger:
Oh, I just remove the error and don't panic at all.  Happy to add that if it would make debugging easier.

PR #438, opened 2024-07-02
Feature: Add functionality of producing hashedNode

tianyueli8227:
When I am creating my own Hashing mechanism based on `go-verkle` library, I notice that there are couple functionality pretty necessary for the library to have but is missing currently. The first one is to create some `HashedNode{}`, and the second one is to get the starting point of starting Hashing Nodes for some given path. 

I have added these two functions as `HashNodeFromInternal` and `GetInternalNode`. Please review.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -1845,3 +1845,40 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
 
 	return result
 }
+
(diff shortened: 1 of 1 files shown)


gballet (commented):
HashedNode is an internal representation that we expect to not make public.

Could you please clarify why you think that these two functions are necessary, and if there is indeed a justification for them, add unit tests?

PR #439, opened 2024-07-08
fix: include non-root node collectNonHashedNodes

tianyueli8227:
In the original implementation of `collectNonHashedNodes`, the function only works when the internal node is root. Since only `BatchSerialize` calls this function by `nodes, paths = n.collectNonHashedNodes(nodes, paths, nil)`, and `BatchSerialize` seems to assume only root will call this function. However, if we want to have any internal node to call `BatchSerialize` without losing generality, we should modify the `collectNonHashedNodes` and make it returns correct leave path correspondingly.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -1728,8 +1728,9 @@ func (n *InternalNode) collectNonHashedNodes(list []VerkleNode, paths [][]byte, 
 	for i, child := range n.children {
 		switch childNode := child.(type) {
 		case *LeafNode:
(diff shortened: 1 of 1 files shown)


gballet:
Thank you for your PR. It is intentional that this function can only be called from the root, and also the change you propose will force more allocations. The point is that we do want to pool all the computation together. Allowing this to be called at lower levels would defeat that purpose. So we won't be merging this.

PR #441, opened 2024-07-10
[wip] SSZ-optimize proof (type 1)

gballet:
This is the first suggestion by Péter, it replaces the proposed suffix diff format with:

```go
type StemStateDiff struct {
	Stem     [StemSize]byte `json:"stem"`
	SuffixDiffs SuffixStateDiffs `json:"suffixDiffs"`
	Suffixes []byte         `json:"suffixes"`
	Current  [][]byte       `json:"current"`
	New      [][]byte       `json:"new"`
}
```

The size improvement compared to the currently proposed method, is significant:

image

The comparison is a bit unfair, since the library doesn't support optionals and so the encoding for the default method is greatly inflated. But one could argue that if optionals can not be implemented, then this is the correct size of the witness.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -30,7 +30,6 @@ import (
 	"errors"
 	"fmt"
 	"sort"
-	"unsafe"
 
 	ipa "github.com/crate-crypto/go-ipa"
 	"github.com/crate-crypto/go-ipa/common"
@@ -83,17 +82,11 @@ type Proof struct {
 	PostValues [][]byte
 }
 
-type SuffixStateDiff struct {
-	Suffix       byte      `json:"suffix"`
-	CurrentValue *[32]byte `json:"currentValue"`
-	NewValue     *[32]byte `json:"newValue"`
-}
-
-type SuffixStateDiffs []SuffixStateDiff
-
 type StemStateDiff struct {
-	Stem        [StemSize]byte   `json:"stem"`
-	SuffixDiffs SuffixStateDiffs `json:"suffixDiffs"`
+	Stem     [StemSize]byte `json:"stem"`
+	Suffixes []byte         `json:"suffixes"`
+	Current  [][]byte       `json:"current"`
+	New      [][]byte       `json:"new"`
 }
 
 type StateDiff []StemStateDiff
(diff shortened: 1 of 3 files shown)


proof_ipa.go
```diff
@@ -83,17 +82,11 @@ type Proof struct {
...
-	Suffix       byte      `json:"suffix"`
-	CurrentValue *[32]byte `json:"currentValue"`
-	NewValue     *[32]byte `json:"newValue"`
-}
-
-type SuffixStateDiffs []SuffixStateDiff
-
 type StemStateDiff struct {
-	Stem        [StemSize]byte   `json:"stem"`
-	SuffixDiffs SuffixStateDiffs `json:"suffixDiffs"`
```
jsign, line 96: OK, now makes sense. The **PR description** has a small mistake that you might want to fix. Delete the `SuffixDiffs SuffixStateDiffs json:"suffixDiffs"` line since this field is removed.

proof_ipa.go
```diff
@@ -83,17 +82,11 @@ type Proof struct {
...
+	Stem     [StemSize]byte `json:"stem"`
+	Suffixes []byte         `json:"suffixes"`
+	Current  [][]byte       `json:"current"`
```
jsign, line 88: We should remember that a proof verifier must now check that `[]byte` has length 32 which was implicit in the type before.

proof_ipa.go
```diff
@@ -102,16 +95,22 @@ func (sd StateDiff) Copy() StateDiff {
...
-		for j := range sd[i].SuffixDiffs {
-			ret[i].SuffixDiffs[j].Suffix = sd[i].SuffixDiffs[j].Suffix
-			if sd[i].SuffixDiffs[j].CurrentValue != nil {
-				ret[i].SuffixDiffs[j].CurrentValue = &[32]byte{}
-				copy((*ret[i].SuffixDiffs[j].CurrentValue)[:], (*sd[i].SuffixDiffs[j].CurrentValue)[:])
+		ret[i].Suffixes = make([]byte, len(sd[i].Suffixes))
+		copy(ret[i].Suffixes, sd[i].Suffixes)
+		ret[i].Current = make([][]byte, len(sd[i].Current))
+		for j := range sd[i].Current {
+			if len(sd[i].Current[j]) == 0 {
```
jsign, line 102: Nit: Remove this empty case, and only leave the else.

proof_ipa.go
```diff
@@ -102,16 +95,22 @@ func (sd StateDiff) Copy() StateDiff {
...
+
+			} else {
+				copy(ret[i].Current[j], sd[i].Current[j])
```
jsign, line 105: I think there's a bug here? We never allocated `ret[i].Current[j]` so it's an empty slice. The `copy` will copy the number of bytes being minimum between  the two parameters, thus 0.

proof_ipa.go
```diff
@@ -102,16 +95,22 @@ func (sd StateDiff) Copy() StateDiff {
...
-				ret[i].SuffixDiffs[j].NewValue = &[32]byte{}
-				copy((*ret[i].SuffixDiffs[j].NewValue)[:], (*sd[i].SuffixDiffs[j].NewValue)[:])
+		}
+		ret[i].New = make([][]byte, len(sd[i].New))
+		for j := range sd[i].New {
+			if len(sd[i].New[j]) == 0 {
+
+			} else {
+				copy(ret[i].New[j], sd[i].New[j])
 			}
```
jsign, line 114: Ditto two comments above.

proof_ipa.go
jsign: Note that in `SerializeProof` an `append` is done, which is correct since it will grow as needed compared to this `copy`.

PR #444, opened 2024-07-22
Move DeserializeAndVerifyVerkleProof to go-verkle

gballet:
This is a request by Péter, that "the trie namespace isn't cluttered by verkle-specific functions. It makes sense to move it here, since it does not depend on any geth declatations.

### DIFF
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -558,7 +558,6 @@ func (x bytesSlice) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
 
 // Verify is the API function that verifies a verkle proofs as found in a block/execution payload.
 func Verify(vp *VerkleProof, preStateRoot []byte, postStateRoot []byte, statediff StateDiff) error {
-	// TODO: check that `OtherStems` have expected length and values.
 
 	proof, err := DeserializeProof(vp, statediff)
 	if err != nil {
--- a/proof_ipa.go
+++ b/proof_ipa.go
@@ -604,9 +604,6 @@ func Verify(vp *VerkleProof, preStateRoot []byte, postStateRoot []byte, statedif
 	if err != nil {
 		return fmt.Errorf("error rebuilding the post-tree from proof: %w", err)
 	}
-	// en fait il crée le blockhash contract, mais il met pas les trucs dans le witness
-	// pourquoi est-ce que c'est triggered maintenant et pourquoi on l'avait pas avant?
-	// hypothèse: en fait ça marchait pas avant non plus mais on teste que le 2ème bloc
 	regeneratedPostTreeRoot := posttree.Commitment().Bytes()
 	if !bytes.Equal(regeneratedPostTreeRoot[:], postStateRoot) {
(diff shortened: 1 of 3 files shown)


proof_ipa.go
```diff
@@ -555,3 +555,62 @@ type bytesSlice []Stem
 func (x bytesSlice) Len() int           { return len(x) }
 func (x bytesSlice) Less(i, j int) bool { return bytes.Compare(x[i], x[j]) < 0 }
 func (x bytesSlice) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
+
+// Verify is the API function that verifies a verkle proofs as found in a block/execution payload.
+func Verify(vp *VerkleProof, preStateRoot []byte, postStateRoot []byte, statediff StateDiff) error {
+	// TODO: check that `OtherStems` have expected length and values.
```
jsign, line 561: I did this TODO a while ago, [here](https://github.com/ethereum/go-verkle/blob/master/proof_ipa.go#L488-L490). I think we can remove this comment.

I also included other [similar checks](https://github.com/ethereum/go-verkle/blob/master/proof_ipa.go#L394-L406), [ordering checks](https://github.com/ethereum/go-verkle/blob/master/proof_ipa.go#L414-L417), and maybe something more I don't remember.

This angle of witness verifications (that I doubt all clients might be doing), should eventually be covered with more focused witness test-vectors. But prob those can't be execution-specs like vectors since can't be generated by "filling" tests from a correctly impl client. Anyway, big tangent here, eventually I can ask Spencer.

proof_ipa.go
```diff
@@ -555,3 +555,62 @@ type bytesSlice []Stem
...
+	fmt.Println(statediff)
```
jsign, line 562: I guess this is being useful now. Pointing out to double-check.

proof_ipa.go
```diff
@@ -555,3 +555,62 @@ type bytesSlice []Stem
...
+	// But all this can be avoided with a even faster way. The EVM block execution can
+	// keep track of the written keys, and compare that list with this post-values list.
+	// This can avoid regenerating the post-tree which is somewhat expensive.
+	posttree, err := PostStateTreeFromStateDiff(pretree, statediff)
+	if err != nil {
+		return fmt.Errorf("error rebuilding the post-tree from proof: %w", err)
+	}
+	// en fait il crée le blockhash contract, mais il met pas les trucs dans le witness
+	// pourquoi est-ce que c'est triggered maintenant et pourquoi on l'avait pas avant?
+	// hypothèse: en fait ça marchait pas avant non plus mais on teste que le 2ème bloc
```
jsign, line 609: Nit: write in English.

jsign (approved):
LGMT, left some comments for your consideration.

proof_ipa.go
gballet: sacrebleu!

gballet, line 609:
```suggestion
```

proof_ipa.go
gballet: nope :)

```suggestion
```

PR #447, opened 2024-07-26
[wip] SSZ-optimize proof (type 3)

jsign:


proof_ipa.go
```diff
@@ -257,32 +292,35 @@ func SerializeProof(proof *Proof) (*VerkleProof, StateDiff, error) {
...
-		case 0:
-			// null value
-		case 32:
-			newsd.CurrentValue = (*[32]byte)(proof.PreValues[i])
+		preLen := len(proof.PreValues[i])
+		postLen := len(proof.PostValues[i])
+		switch {
+		case isInsertion(preLen, postLen):
+			var newValue [32]byte
+			copy(newValue[:], proof.PostValues[i])
```
jsign, line 300: Note that we don't have to do the `postLen == 32` and all the alignment thing.

If `proof.PostValues[i]` isn't 32 bytes, then `copy` will copy whatever amount of bytes there are. So both in aligned and unaligned case, the right thing will happen.

Same applies to other cases.

proof_ipa.go
```diff
@@ -524,20 +577,20 @@ func PostStateTreeFromStateDiff(preroot VerkleNode, statediff StateDiff) (Verkle
...
 	for _, stemstatediff := range statediff {
 		var (
-			values     = make([][]byte, NodeWidth)
-			overwrites bool
+			values    = make([][]byte, NodeWidth)
+			overwrite bool
 		)
 
-		for _, suffixdiff := range stemstatediff.SuffixDiffs {
-			if /* len(suffixdiff.NewValue) > 0 - this only works for a slice */ suffixdiff.NewValue != nil {
```
jsign, line 532: This previous `if` (that was kept in Type 2) was removed. This can also be removed in Type 2 as I mentioned in my review in that PR.

Both in `Inserts` and `Updates`, we can't have "nil" or empty values in `New`.

proof_json.go
```diff
@@ -183,32 +184,46 @@ func (vp *VerkleProof) UnmarshalJSON(data []byte) error {
...
 
 func (ssd StemStateDiff) MarshalJSON() ([]byte, error) {
+	suffixes := make([]byte, 0, len(ssd.Updates)+len(ssd.Inserts))
+	for i := range ssd.Updates {
+		suffixes = append(suffixes, ssd.Updates[i].Suffix)
+	}
+	for i := range ssd.Inserts {
+		suffixes = append(suffixes, ssd.Inserts[i].Suffix)
+	}
+	sort.Slice(suffixes, func(i, j int) bool { return suffixes[i] < suffixes[j] })
```
jsign, line 203: I think in [Type 2 there might be a bug](https://github.com/ethereum/go-verkle/pull/443/files#diff-bd86229874add74bf1b46ffcac89ca1157aa411a771321037ba2c63f84989214R197) since it's only inserting Updates, but prob we need both Updates+Inserts as I did here?
I also sort them since I'd guess that's expected.

proof_ipa.go
```diff
@@ -102,18 +117,28 @@ func (sd StateDiff) Copy() StateDiff {
...
-			}
-			if sd[i].SuffixDiffs[j].NewValue != nil {
-				ret[i].SuffixDiffs[j].NewValue = &[32]byte{}
-				copy((*ret[i].SuffixDiffs[j].NewValue)[:], (*sd[i].SuffixDiffs[j].NewValue)[:])
-			}
+
+		ret[i].Updates = make([]UpdateDiff, len(sd[i].Updates))
+		for j := range sd[i].Updates {
+			ret[i].Updates[j].Suffix = sd[i].Updates[j].Suffix
+			ret[i].Updates[j].Current = sd[i].Updates[j].Current
```
jsign, line 124: Since we stop having `[]byte`, many of `if len(xxx) == 0` cases and similar are gone. We can always copy-by-value.

proof_json.go
gballet: that's possible, I'm only focusing on the SSZ encoding at the moment. We'll pick the best method and fix it if need be.

Issue #448, opened 2024-08-20
deletion of a non-existent account overwrites with 0s


gballet:
In geth, there is an upcoming branch in which an account can be deleted if it doesn't exist. An actual call to deletion is made, which means that the account is ovewritten with 0s. Now, it never happened before since a non-existent account would not be "deleted", it would just be ignored. But it seems that in [this new branch[(https://github.com/rjl493456442/go-ethereum/pull/12), an actual deletion command is sent.

This raises the question, independent of this geth branch: if an account is deleted, that doesn't exist in the tree, should it be "created" and then overwritten with 0s, or just ignored ?

gballet:
Writing down here after a talk with @jsign : this stuff is only needed for:

 1) this PR by Gary, which honestly sounds like something that should be handled at the client level and probably can be in `trie/verkle.go`.
 2) replays, which is the more tricky thing. It should be considered whether pure deletion can be implemented during replays. Probably not, owing to the impossibility of deleting a contract storage in verkle, but a clever trick might simplify all that.

PR #449, opened 2024-08-21
Add leaf group "erasure" function

gballet:
This is needed to fix tests in https://github.com/rjl493456442/go-ethereum/pull/12 : Gary's stuff will call `DeleteAccount` if the account is non-existent. While this is quite unclean and should be fixed at the geth layer, I reckon that it is much more difficult to do than designing the verkle library to handle this.

As a result, I introduce an `EraseStem` method, which is used to replace `InsertAtStem` when deleting a function: it will go through the tree, find a group if it exsits, and overwrite all values present in that group. This is less invasive than the current method of calling `InsertAtStem` with 256 non-zero values.

It has the following drawbacks:
 * This behavior needs to be specified in the spec, since `nil` values are not overwritten
 * It also breaks the replay code, which uses the trick of checking that all values are 0 in order to simulate the "pre-verge" behavior of account deletion.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -631,6 +631,53 @@ func (n *InternalNode) Delete(key []byte, resolver NodeResolverFn) (bool, error)
 	}
 }
 
+// EraseStem overwrites values with 0s, which is the (currently specced) way of deleting
+// leaves in verkle trees. It will, however, not create anything if it finds that the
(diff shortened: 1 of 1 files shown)


gballet:
Fixes #448

PR #451, opened 2024-08-28
ci: improvements

jsign:
This PR improves the CI configuration to:
- Remove workflows that now have native support in GH through Settings.
- Stop using `self-hosted` for build/lint/test. It isn't properly configured and took ~30min to run for no good reason (i.e: has an ever growing Go cache restoring).
- The new build/test covers three OS.


.github/workflows/go.yml
```diff
@@ -1,46 +1,54 @@
...
-      uses: actions/setup-go@v4
-      with:
-        go-version: 1.18
-    - name: Build
-      run: go build -v ./...
+      - uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22
```
jsign, line 21: Targeting the same version as the minimum [required today in go-ethereum](https://github.com/ethereum/go-ethereum/blob/master/go.mod#L5).

.github/workflows/go.yml
```diff
@@ -1,46 +1,52 @@
...
+        with:
+          version: v1.60
+      - name: Vet
+        run: go vet
+
   test:
-    runs-on: self-hosted
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
jsign, line 43: Running tests in three OSs.

.github/workflows/go.yml
```diff
@@ -1,46 +1,52 @@
...
-        go-version: 1.18
-    - name: Test
-      run: go test -v -race ./...
+      - uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22
+      - name: Test
+        run: go test -v -race ./... -timeout=15m
```
jsign, line 52: Looks like the `macos-latest` is a bit slow, so it requires a bit more time.

jsign: Uhm, last run took 7m. But I saw at least one take >10min thus failing.
Let's keep 15m since maybe GH have different machines -- let's avoid potential false positives.

tree_test.go
```diff
@@ -33,7 +33,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	mRand "math/rand"
+	mRandV1 "math/rand"
```
gballet, line 36: can we not convert everything to v2?

.github/workflows/codeql.yml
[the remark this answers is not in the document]
gballet: setting them in settings will commit these files, this is how they ended up in this repo. They should remain. If I follow the instructions, it tells me that enabling it will override the "advanced" setup.

image

gballet (changes requested):
Regarding `selfhosted`, we are currently looking into fixing this problem, which needs to be solved for geth as well. So I don't think there is a need to move away from it, especially since `ubuntu-latest` is not powerful enough to run all the geth tests.

tree_test.go
jsign: Unfortuantely no until we can use Go v1.23 which creates an API to do that exactly.

.github/workflows/codeql.yml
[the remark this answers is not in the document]
jsign: I'll restore these files then.

jsign:
> Regarding `selfhosted`, we are currently looking into fixing this problem, which needs to be solved for geth as well. So I don't think there is a need to move away from it, especially since `ubuntu-latest` is not powerful enough to run all the geth tests.

Why is `ubuntu-latest` power related to `go-verkle`? geth tests can still be run with `self-hosted` if that's useful.  What's the benefit of running `self-hosted` in this repo?

gballet (approved):

PR #454, opened 2024-09-11
Use skiplists to save on the bitmap size when compressing leaves

gballet:
Instead of using a bitmap, use two bytes:

 * the first byte will indicate the length of the range of empty leaves
 * the second byte will indicate the length of the next range of non-empty leaves

There are ~200M accounts in the state, for which we use at most 6 bytes instead of 32 for a bitmap. This is expected to save (32 - 6) * 200 ~ 5GB

There are further optimizations that will not be handled by this PR, but that I'm writing down here so that I don't forget:

 * left-trim leaves
 * only store the C1/C2 if necessary (I need to find a bit to flag which ones are not present)

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -56,12 +56,11 @@ const (
 	leafC1CommitmentOffset = leafCommitmentOffset + banderwagon.UncompressedSize
 	leafC2CommitmentOffset = leafC1CommitmentOffset + banderwagon.UncompressedSize
 	leafChildrenOffset     = leafC2CommitmentOffset + banderwagon.UncompressedSize
-	leafBalanceSize        = 32
-	leafNonceSize          = 8
+	leafBasicDataSize      = 32
 	leafSlotSize           = 32
 	leafValueIndexSize     = 1
 	singleSlotLeafSize     = nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafValueIndexSize + leafSlotSize
-	eoaLeafSize            = nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafBalanceSize + leafNonceSize
+	eoaLeafSize            = nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafBasicDataSize
 )
 
 func bit(bitlist []byte, nr int) bool {
@@ -94,6 +93,8 @@ func ParseNode(serializedNode []byte, depth byte) (VerkleNode, error) {
 		return parseEoAccountNode(serializedNode, depth)
 	case singleSlotType:
 		return parseSingleSlotNode(serializedNode, depth)
+	case skipListType:
+		return parseSkipList(serializedNode, depth)
 	default:
 		return nil, ErrInvalidNodeEncoding
 	}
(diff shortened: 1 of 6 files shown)


tree.go
```diff
@@ -170,6 +170,7 @@ const (
 	leafType       byte = 2
 	eoAccountType  byte = 3
 	singleSlotType byte = 4
+	skipListType   byte = 8
```
gballet, line 173: this is set in bit order, because I'm hoping to mix encodings in the future.

tree.go
```diff
@@ -1775,35 +1776,42 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
...
-				// Nonce should have its last 24 bytes set to 0
-				isEoA = v != nil && bytes.Equal(v[leafNonceSize:], zero24[:])
-			case 3:
+			case 1:
 				// Code hash should be the empty code hash
 				isEoA = v != nil && bytes.Equal(v, EmptyCodeHash[:])
-			case 4:
-				// Code size must be 0
-				isEoA = v != nil && bytes.Equal(v, zero32[:])
 			default:
```
gballet, line 1815: This is the EoA fix for Nyota.

proof_test.go
gballet, line 1: I believe that the changes in this PR are due to me building on top of two of your commits, which I need for my replay experiments.

proof_ipa.go
```diff
@@ -347,22 +386,37 @@ func DeserializeProof(vp *VerkleProof, statediff StateDiff) (*Proof, error) {
...
+			prevalues = append(prevalues, upt.Current[:])
+			postvalues = append(postvalues, upt.New[:])
+		}
+		for _, ins := range stemdiff.Inserts {
+			var k [32]byte
+			copy(k[:StemSize], stemdiff.Stem[:])
+			k[StemSize] = ins.Suffix
+			keys = append(keys, k[:])
+			prevalues = append(prevalues, nil)
+			postvalues = append(postvalues, slices.Clone(ins.New[:]))
```
gballet, line 403: This, and line 410, are the fixes for the test that was broken.

jsign: Weird, so this means that `stemdiff` is somewhat mutated after affecting the returned `Proof`?

tree.go
jsign: `eoAccountType` has value 3 which isn't respecting that idea. Is it worth changing now? If not, we're ina a half-baked idea.

tree.go
```diff
@@ -1830,8 +1838,28 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
...
+		result[0] = skipListType
+		result = append(result, n.stem[:StemSize]...)
+		result = append(result, cBytes[:]...)
+		result = append(result, c1Bytes[:]...)
+		result = append(result, c2Bytes[:]...)
+		var leafIdx int
+		for _, gap := range gaps {
+			if gap.Count == 0 {
+				break // skip the last gap as nothing follows
+			}
```
jsign, line 1854: Could we do:
```
for j:=0; j<=gapCount; j++ {
```
and avoid `if` L1852 ?

encoding.go
```diff
@@ -135,17 +136,49 @@ func parseLeafNode(serialized []byte, depth byte) (VerkleNode, error) {
 	return ln, nil
 }
 
+func parseSkipList(serialized []byte, depth byte) (VerkleNode, error) {
+	var values [NodeWidth][]byte
+	offset := leafStemOffset + StemSize + 3*banderwagon.UncompressedSize // offset in the serialized payload
+	valueIdx := 0                                                        // Index of the value being deserialized
+	for valueIdx < NodeWidth {
+		rangecount := serialized[offset+1]
+		gapsize := serialized[offset]
```
jsign, line 145: Can we rename `gapsize` and `valueIdx` to `rangeSkip` and `rangeCount`, as to use the same names in the serialization `Skip` and `Count`?

Using `gap` or `range` worlds can be confusing. Using the same names as the serialization makes easier to understand the both ways of the algorithm.

jsign (approved):
LGTM, left some comments for your consideration.

proof_ipa.go
gballet: the area that is used to store the location of the iterator is reused over and over, and if we pass a reference to it, it will change as the iterator progress. This is not the first time we fall for this, and probably not the last time either, it's quite subtle and easy to forget.

tree.go
gballet: we could change it to 4, no long-term db has been using this.

tree.go
gballet: not completely, because if `Count == 0` it means that the group ends with a gap, wherease if `Count != 0` then the group ends with a range. But yeah, no point in going over the whole list, I'll loop on `j`.

tree.go
```diff
@@ -1775,35 +1776,42 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
 		bitlist        [bitlistSize]byte
 		isEoA          = true
 		count, lastIdx int
+		gapcount       int
+		gaps           [32]struct {
+			Skip  byte // How many slots to skip before the next range
+			Count int  // Size of the next range
```
gballet, line 1782:
```suggestion
			Count int  // Size of the next range. `int` because a full leaf has 256 entries
```

tree.go
```diff
@@ -1830,8 +1838,31 @@ func (n *LeafNode) serializeLeafWithUncompressedCommitments(cBytes, c1Bytes, c2B
...
+		result = append(result, c1Bytes[:]...)
+		result = append(result, c2Bytes[:]...)
+		var leafIdx int
+		for _, gap := range gaps {
+			if gap.Count == 0 {
+				break // skip the last gap as nothing follows
+			}
+			result = append(result, gap.Skip)
+			leafIdx += int(gap.Skip)
+			result = append(result, byte(gap.Count))
```
gballet, line 1857: I'm thinking there can't be a 0-length range, so we should encode `Count-1` instead of forcing the wraparound, which might not work on every platform.

```suggestion
			leafIdx += int(gap.Skip)
			// count can be 256, bytes(256) == 0 which is used as  a marker for a full leaf
			result = append(result, byte(gap.Count))
```

encoding.go
```diff
@@ -56,6 +56,9 @@ const (
 	leafC1CommitmentOffset = leafCommitmentOffset + banderwagon.UncompressedSize
 	leafC2CommitmentOffset = leafC1CommitmentOffset + banderwagon.UncompressedSize
 	leafChildrenOffset     = leafC2CommitmentOffset + banderwagon.UncompressedSize
+	leafSkipListCOffset    = nodeTypeSize + StemSize
```
jsign, line 59: Nit: `leafSkipListC0Offset`

encoding.go
```diff
@@ -140,35 +143,44 @@ func parseSkipList(serialized []byte, depth byte) (VerkleNode, error) {
...
-	for valueIdx < NodeWidth {
-		rangecount := serialized[offset+1]
-		gapsize := serialized[offset]
-		valueIdx += int(gapsize)
-		offset += 2
-		for i := 0; i < int(rangecount); i++ {
-			values[valueIdx] = serialized[offset : offset+leafSlotSize]
+
+	// shortcut: the leaf is full and so both values are 0.
+	if serialized[offset] == 0 && serialized[offset+1] == 0 {
```
jsign, line 148: Is this worth it? If the node is full, doesn't sound we should serialize with skip list really. Sounds wasteful.

PR #457, opened 2024-10-11
Add EIP7736 support 

weiihann:
Primary changes:
- Add new `StateEpoch` type
- Modify `Get`, `Insert` and `Delete` to include `StateEpoch`. If trying to access expired leaf node, returns error.
- Additional `lastEpoch` field in `LeafNode`, included in serialization and node commitment calculation
- Add new `ExpiryLeafNode` struct

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -13,13 +13,14 @@ import (
 
 // BatchNewLeafNodeData is a struct that contains the data needed to create a new leaf node.
 type BatchNewLeafNodeData struct {
-	Stem   Stem
-	Values map[byte][]byte
+	Stem      Stem
+	Values    map[byte][]byte
+	LastEpoch StateEpoch
 }
 
 // BatchNewLeafNode creates a new leaf node from the given data. It optimizes LeafNode creation
 // by batching expensive cryptography operations. It returns the LeafNodes sorted by stem.
-func BatchNewLeafNode(nodesValues []BatchNewLeafNodeData) ([]LeafNode, error) {
+func BatchNewLeafNode(nodesValues []BatchNewLeafNodeData, curEpoch StateEpoch) ([]LeafNode, error) {
 	cfg := GetConfig()
 	ret := make([]LeafNode, len(nodesValues))
 
@@ -45,7 +46,7 @@ func BatchNewLeafNode(nodesValues []BatchNewLeafNodeData) ([]LeafNode, error) {
 					}
 
 					var leaf *LeafNode
-					leaf, err := NewLeafNode(nv.Stem, valsslice)
+					leaf, err := NewLeafNode(nv.Stem, valsslice, nv.LastEpoch)
 					if err != nil {
 						return err
 					}
@@ -97,14 +98,14 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
 	panic("stems are equal")
 }
 
-func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
+func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, curEpoch StateEpoch, resolver NodeResolverFn) error {
 	sort.Slice(leaves, func(i, j int) bool {
 		return bytes.Compare(leaves[i].stem, leaves[j].stem) < 0
 	})
 
 	// We first mark all children of the subtreess that we'll update in parallel,
 	// so the subtree updating doesn't produce a concurrent access to n.cowChild(...).
-	var lastChildrenIdx = -1
+	lastChildrenIdx := -1
 	for i := range leaves {
 		if int(leaves[i].stem[0]) != lastChildrenIdx {
 			lastChildrenIdx = int(leaves[i].stem[0])
@@ -132,13 +133,13 @@ func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeReso
 			start := currStemFirstByte
 			end := i
 			group.Go(func() error {
-				return n.insertMigratedLeavesSubtree(leaves[start:end], resolver)
+				return n.insertMigratedLeavesSubtree(leaves[start:end], curEpoch, resolver)
 			})
 			currStemFirstByte = i
 		}
 	}
 	group.Go(func() error {
-		return n.insertMigratedLeavesSubtree(leaves[currStemFirstByte:], resolver)
+		return n.insertMigratedLeavesSubtree(leaves[currStemFirstByte:], curEpoch, resolver)
 	})
 	if err := group.Wait(); err != nil {
 		return fmt.Errorf("inserting migrated leaves: %w", err)
@@ -147,7 +148,7 @@ func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeReso
 	return nil
 }
 
-func (n *InternalNode) insertMigratedLeavesSubtree(leaves []LeafNode, resolver NodeResolverFn) error { // skipcq: GO-R1005
+func (n *InternalNode) insertMigratedLeavesSubtree(leaves []LeafNode, curEpoch StateEpoch, resolver NodeResolverFn) error { // skipcq: GO-R1005
 	for i := range leaves {
 		ln := leaves[i]
 		parent := n
@@ -192,7 +193,7 @@ func (n *InternalNode) insertMigratedLeavesSubtree(leaves []LeafNode, resolver N
 					}
 				}
 
-				if err := node.updateMultipleLeaves(nonPresentValues); err != nil {
+				if err := node.updateMultipleLeaves(nonPresentValues, curEpoch); err != nil {
 					return fmt.Errorf("updating leaves: %s", err)
 				}
 				continue
--- a/encoding.go
+++ b/encoding.go
@@ -56,12 +56,13 @@ const (
 	leafC1CommitmentOffset = leafCommitmentOffset + banderwagon.UncompressedSize
 	leafC2CommitmentOffset = leafC1CommitmentOffset + banderwagon.UncompressedSize
 	leafChildrenOffset     = leafC2CommitmentOffset + banderwagon.UncompressedSize
+	leafLastEpochOffset    = leafChildrenOffset + EpochSize
 	leafBalanceSize        = 32
 	leafNonceSize          = 8
 	leafSlotSize           = 32
 	leafValueIndexSize     = 1
-	singleSlotLeafSize     = nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafValueIndexSize + leafSlotSize
-	eoaLeafSize            = nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafBalanceSize + leafNonceSize
+	singleSlotLeafSize     = nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafValueIndexSize + leafSlotSize + EpochSize
+	eoaLeafSize            = nodeTypeSize + StemSize + 2*banderwagon.UncompressedSize + leafBalanceSize + leafNonceSize + EpochSize
 )
 
 func bit(bitlist []byte, nr int) bool {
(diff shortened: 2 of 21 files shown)


expired_leaf.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"errors"
+)
+
+type ExpiredLeafNode struct {
+	stem       Stem
+	commitment *Point
+}
+
+var errExpiredLeafNode = errors.New("trying to access an expired leaf node")
```
gballet, line 37: this is redundant with `errEpochExpired` in `doc.go`, I believe.

expired_leaf.go
```diff
@@ -0,0 +1,102 @@
...
+	return n.commitment
+}
+
+func (n ExpiredLeafNode) Commitment() *Point {
+	return n.commitment
+}
+
+func (n ExpiredLeafNode) GetProofItems(keylist, NodeResolverFn) (*ProofElements, []byte, []Stem, error) {
+	return nil, nil, nil, errExpiredLeafNode
+}
```
gballet, line 64: I don't think that this is correct: you need to be able to prove that something was expired, for the block to be able to execute statelessly.

We might have to change the proof format, to add "expired leaves" like we have "proof-of-absence" stems.

state_epoch.go
```diff
@@ -0,0 +1,27 @@
...
+	"encoding/binary"
+)
+
+type StateEpoch uint64
+
+const (
+	NumActiveEpochs = 2
+)
+
+func EpochExpired(prev StateEpoch, cur StateEpoch) bool {
```
gballet, line 13: Prefer `IsExpired`

state_epoch.go
```diff
@@ -0,0 +1,27 @@
...
+	return cur-prev >= NumActiveEpochs
+}
+
+// Convert epoch to bytes
+func (e StateEpoch) Bytes() []byte {
+	b := make([]byte, 8)
+	binary.BigEndian.PutUint64(b, uint64(e))
+	return b
+}
```
gballet, line 22: I would very much like not to have this method, it introduces an extra allocation+copy when this is used.

state_epoch.go
gballet, line 1: Epoch is a bad name. First because it clashes with the CL concept of epoch, and second because it isn't an epoch: it's an access timestamp. Unless you decided to use actual epochs there? If so, that might work, but it will cause the db to be cleared at the end of an epoch, which isn't as efficient as having some sort of garbage collection process, that slowly runs to clear the db when there is time.

proof_ipa.go
```diff
@@ -145,7 +145,7 @@ func getProofElementsFromTree(preroot, postroot VerkleNode, keys [][]byte, resol
 		// keys were sorted already in the above GetcommitmentsForMultiproof.
 		// Set the post values, if they are untouched, leave them `nil`
 		for i := range keys {
-			val, err := postroot.Get(keys[i], resolver)
+			val, err := postroot.Get(keys[i], 0, resolver)
```
gballet, line 148: I don't think this is correct: if the epoch is 0 here, then you would get `errEpochExpired` all the time, wouldn't you?

gballet (commented):
Thank you for this PR. This is a good start, although it requires a lot of polishing, and probably a couple spec discussions as well. I left a few comments for you to consider, I can have another pass when you've had a look.

state_epoch.go
weiihann: Access timestamp is probably easier to understand, but I don't think it's exactly a "timestamp" either. I guess it's more like a counter, where we increment it for every `EPOCH_LENGTH`.

Regardless I'll rename it to access timestamp first, until we decided a better naming for it and change it in the EIP.

weiihann:
Here's a general explanation of how expiry works with stateless proofs:

There are 4 scenarios for a given stem:
1. Both prestate and poststate not expired
- Works as normal

2. Both prestate and poststate expired
- Works as normal, there shouldn't be any state diffs because they are expired

3. Prestate not expired, poststate expired
- Works as normal, this indicates that there are no state updates from prestate to poststate, so state diffs do not exist

4. Prestate expired, poststate not expired
- This indicates that a resurrection has occurred. How to verify:
  - We added a new extension status type called `extStatusExpired` to indicate if a given stem has expired.
  - If an extension status of a stem is `extStatusExpired` and any of the postvalue in the stem given in the proof is not absent, then we know that there's a resurrection.

proof_ipa.go
```diff
@@ -94,6 +94,7 @@ type SuffixStateDiffs []SuffixStateDiff
 type StemStateDiff struct {
 	Stem        [StemSize]byte   `json:"stem"`
 	SuffixDiffs SuffixStateDiffs `json:"suffixDiffs"`
+	Resurrected bool             `json:"resurrected"`
```
gballet, line 97: This is going to be taking more space. I would do a `expiredStem` field like we have `poaStems`. But no need to take action on this right now, as long as it works it's fine. This is a later-stage optimization.

encoding.go
```diff
@@ -163,12 +172,13 @@ func parseEoAccountNode(serialized []byte, depth byte) (VerkleNode, error) {
 func parseSingleSlotNode(serialized []byte, depth byte) (VerkleNode, error) {
 	var values [NodeWidth][]byte
 	offset := leafStemOffset
-	ln := NewLeafNodeWithNoComms(serialized[offset:offset+StemSize], values[:])
+	epochOffset := leafStemOffset + StemSize + 2*banderwagon.UncompressedSize
+	ln := NewLeafNodeWithNoComms(serialized[offset:offset+StemSize], values[:], AccessTimestampFromBytes(serialized[epochOffset:epochOffset+EpochSize]))
```
gballet, line 176: another reason not to do timestamp after all, is that timestamps take more disk space than an epoch, which could be encoded as two bytes for a very, very long time (if an epoch is 6 months, that would be 32000+ years), which is good for disk space since all leaf nodes will have to have an epoch.

With a timestamp, that would mean 4 bytes (as of today) and will need another byte by 2106 (still far out, but not as good)

tree.go
```diff
@@ -1503,10 +1589,23 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
 		}
 
 		esses []byte = nil // list of extension statuses
-		poass []Stem       // list of proof-of-absence stems
+		poass []Stem       // list of proof-of-absence and proof-of-expiry stems
 	)
 
-	// Initialize the top-level polynomial with 1 + stem + C1 + C2
+	// If the leaf node is expired, generate the proof of expiry.
+	if IsExpired(n.lastTs, curTs) {
```
gballet, line 1596: I don't think we need to change the API for `GetProofItems` to have a timestamp/epoch: this is because when we do the proof, the relevant elements of the tree have already been loaded with `Get`, which is where the detection can happen. It would be simpler to catch things there, and assume that, at proof creation time, if a node is not expired, then it's not expired in the db either.

tree.go
```diff
@@ -1244,13 +1309,14 @@ func (n *LeafNode) updateLeaf(index byte, value []byte) error {
 
 	// If index is in the first NodeWidth/2 elements, we need to update C1. Otherwise, C2.
 	cxIndex := 2 + int(index)/(NodeWidth/2) // [1, stem, -> C1, C2 <-]
-	n.updateC(cxIndex, frs[0], frs[1])
+	n.updateC(cxIndex, frs[0], frs[1], curTs)
+	n.updateLastTs(curTs)
```
gballet, line 1313: `updateLastTs` doesn't update the commitment, so it should be called _before_ `updateC`

tree.go
```diff
@@ -1554,12 +1653,21 @@ func (n *LeafNode) GetProofItems(keys keylist, _ NodeResolverFn) (*ProofElements
 		pe.Fis = append(pe.Fis, poly[:])
 	}
 
+	// add last accessed timestamp
+	poly[4].SetUint64(uint64(n.lastTs))
```
gballet, line 1657:
```suggestion
	poly[4].SetUint64(uint64(n.lastTs))

```

gballet (commented):
Left more change requests

conversion.go
```diff
@@ -97,14 +97,14 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
...
 
-func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
+func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, curTs AccessTimestamp, resolver NodeResolverFn) error {
 	sort.Slice(leaves, func(i, j int) bool {
 		return bytes.Compare(leaves[i].stem, leaves[j].stem) < 0
 	})
 
 	// We first mark all children of the subtreess that we'll update in parallel,
 	// so the subtree updating doesn't produce a concurrent access to n.cowChild(...).
-	var lastChildrenIdx = -1
```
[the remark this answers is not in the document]
weiihann: It's the linter 😵‍💫 Will revert those changes

conversion.go
```diff
@@ -97,14 +97,14 @@ func firstDiffByteIdx(stem1 []byte, stem2 []byte) int {
 	panic("stems are equal")
 }
 
-func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, resolver NodeResolverFn) error {
+func (n *InternalNode) InsertMigratedLeaves(leaves []LeafNode, curTs AccessTimestamp, resolver NodeResolverFn) error {
```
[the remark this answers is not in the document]
weiihann: No worries, I prefer epochs too :)

expired_tree_test.go
```diff
@@ -0,0 +1,243 @@
...
+		t.Fatalf("expected value %x, got %x", testValue, leaff.values[ffx32KeyTest[StemSize]])
+	}
+
+	if leaf0.lastPeriod != 0 {
+		t.Fatalf("expected last accessed to be 0, got %d", leaf0.lastPeriod)
+	}
+
+	if leaff.lastPeriod != 2 {
+		t.Fatalf("expected last accessed to be 2, got %d", leaff.lastPeriod)
+	}
```
gballet, line 102: you should also try to insert a sibling into an expired leaf, I might have missed it, but it seems to me that this case isn't being tested.

e.g. insert(00...0000, 0), and then insert(00..1100, 2) and see if intermediate branch nodes are inserted.

gballet (commented):
* This PR must be updated to refresh the period upon a read
* Also, I don't think that the tests are actually checking that an expiration has been proven. but maybe I misundertsood.
* It also seems to me that inserting a sibling to an expired node isn't tested.

epoch.go
```diff
@@ -0,0 +1,19 @@
+package verkle
+
+import (
+	"encoding/binary"
+)
+
+type StatePeriod uint16
```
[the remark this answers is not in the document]
weiihann: As per our convo, this is a premature optimization and this comment should be discarded. We can look into it in the future.

weiihann:
Change Log (5/1/2025):
- Add `Revive()` to `VerkleNode` interface
- `Proof` now includes the periods of each stem in pre-state tree. How stateless proof works with expiry:
  - Given a pre-state tree with expired stem, we first perform `Revive()`
  - This particular `Revive()` operation should not refresh the period of the leaf node to prevent changing commitment
  - We now have a pre-state tree with expired stems resolved
  - Generate proof. For each stem, the period is included in the proof
  - In the verification process, the pre-state tree can be reconstructed from the proof, and the period of each stem is resolved correctly
- Minor code quality improvements

tree.go
```diff
@@ -432,14 +473,68 @@ func (n *InternalNode) InsertValuesAtStem(stem Stem, values [][]byte, resolver N
...
+		leaf.setDepth(n.depth + 1)
+		if !skipUpdate {
+			leaf.updatePeriod(curPeriod)
+		}
+		n.children[nChild] = leaf
+		n.cowChild(nChild)
+		
+		return nil
+	case *LeafNode:
+		return child.Revive(stem, values, oldPeriod, curPeriod, skipUpdate, resolver)
```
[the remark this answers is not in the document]
gballet: I get it now, the field might not have been pruned by the collector. Add a comment here, so that I don't ask myself that question each time, please.

tree.go
```diff
@@ -90,14 +91,16 @@ type VerkleNode interface {
...
-	GetProofItems(keylist, NodeResolverFn) (*ProofElements, []byte, []Stem, error)
+	GetProofItems(keylist, NodeResolverFn) (*ProofElements, []byte, []Stem, []StatePeriod, error)
 
 	// Serialize encodes the node to RLP.
 	Serialize() ([]byte, error)
 
 	// Copy a node and its children
 	Copy() VerkleNode
 
+	Revive(Stem, [][]byte, StatePeriod, StatePeriod, bool, NodeResolverFn) error
```
gballet, line 102: could you please document what the boolean does?

gballet (commented):
It's getting into shape! I'm not sure that i understand the point of `skipUpdate`: why shouldn't the counter/commitment be updated upon revival? We can think of a use case where someone resurects an account for someone else, and doesn't directly write/read from that account.

tree.go
```diff
@@ -432,14 +473,68 @@ func (n *InternalNode) InsertValuesAtStem(stem Stem, values [][]byte, resolver N
...
+	}
+
+	nChild := offset2key(stem, n.depth)
+
+	switch child := n.children[nChild].(type) {
+	case UnknownNode:
+		return errMissingNodeInStateless
+	case Empty:
+		// TODO(weiihann): double confirm this
+		return errors.New("cannot revive an empty node")
```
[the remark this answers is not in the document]
weiihann: I'll just leave it as error for now and add the scenario you mentioned in the comments

weiihann:
> I'm not sure that i understand the point of skipUpdate: why shouldn't the counter/commitment be updated upon revival?
It's because of stateless proof. If the pre-state tree contains expired nodes, it has to be revived first. If the counter is updated directly, the commitment of the node and the tree changes. Pre-state tree has to be revived such that it preserves the original values and counter, so that it can generate the proof items properly.

Though there might be better ways of doing this. Another alternative is to cache the previous counter and previous commitment's information, but that makes the logic more complex. I would have to look at how the stateless proof is generated in geth and revisit this if needed.

PR #458, opened 2024-10-21
add proof Equal methods

jsign:
This PR adds `.Equal` methods to proof fields so clients can compare re-generated proofs against block proofs.

This is required for https://github.com/gballet/go-ethereum/pull/515

### DIFF
--- a/proof_json.go
+++ b/proof_json.go
@@ -42,6 +42,10 @@ func PrefixedHexStringToBytes(input string) ([]byte, error) {
 	if input[0:2] == "0x" {
 		input = input[2:]
 	}
+	// TODO(jsign): workaround for execution-spec-tests bug. Remove as soon as it's fixed.
+	if len(input)%2 != 0 {
+		input = "0" + input
+	}
 	return hex.DecodeString(input)
 }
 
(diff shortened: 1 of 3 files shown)


proof_json.go
```diff
@@ -42,6 +42,10 @@ func PrefixedHexStringToBytes(input string) ([]byte, error) {
 	if input[0:2] == "0x" {
 		input = input[2:]
 	}
+	// TODO(jsign): workaround for execution-spec-tests bug. Remove as soon as it's fixed.
+	if len(input)%2 != 0 {
+		input = "0" + input
+	}
```
jsign, line 48: This is a workaround to deal with a situation where `execution-spec-tests` do an unmarshaling and marshaling of the execution witness and produce odd-length hex values (when they weren't present in the original witness generated by geth). The `hex` stdlib take odd-length hex as invalid.

I'll talk with Spencer to fix this on the testing library. Meanwhile, I'm doing this change to deal with the situation and be able to unmarshal `execution-spec-tests` provided JSON witnesses.

I added a TODO to remove this since we should be as strict as possible in accepting/rejecting serialized witnesses. In anycase, JSON deserialization will probably disappear or at a minimum not be used for mainnet flows, since we'll use SSZ.

gballet (approved):
LGTM. It's a bit annoying that we'll have to do it again when we switch to type 3 encoding, but we'll manage I guess.

PR #462, opened 2025-01-06
Verify: only check post root if provided

gballet:
Don't try to verify the validity of the provided post-state root if it isn't provided.

This is necessary for gballet/go-ethereum#527 to work.


jsign (approved):
LGTM.

I see that the tests fail, is this related to this bug still or another?

gballet:
It's related to a bug we introduced when disabling the eoa optimization. I am about to push a fix.

PR #467, opened 2025-06-16
feat: Add error handling to Commit() function

Apostlex0:
### Description
This PR adds error handling to the `Commit()` function in the Verkle tree implementation. Previously, the `Commit()` function did not return errors, which could lead to silent failures. 

### Changes
- Modified the `VerkleNode` interface to make `Commit()` return `(*Point, error)`
- Updated all implementations of `Commit()` to properly handle and return errors
- Updated all callers of `Commit()` to handle the returned error
- Fixed test cases to properly handle errors from `Commit()`

### Why
The original implementation did not handle potential errors during commitment computation, which could lead to:
1. Silent failures in critical operations
2. Inconsistent state when errors occur
3. Difficulty in debugging issues
4. Potential security implications

### Testing
- All tests have been updated to handle errors from `Commit()` and they pass

### Related Issues
- Improves error handling consistency across the codebase

### DIFF
--- a/hashednode.go
+++ b/hashednode.go
@@ -44,14 +44,8 @@ func (HashedNode) Get([]byte, NodeResolverFn) ([]byte, error) {
 	return nil, errors.New("can not read from a hash node")
 }
 
-func (HashedNode) Commit() *Point {
-	// TODO: we should reconsider what to do with the VerkleNode interface and how
-	//       HashedNode fits into the picture, since Commit(), Commitment() and Hash()
-	//	     now panics. Despite these calls must not happen at runtime, it is still
-	//	     quite risky. The reason we end up in this place is because PBSS came quite
-	//	     recently compared with the VerkleNode interface design. We should probably
-	//	     reconsider splitting the interface or find some safer workaround.
-	panic("can not commit a hash node")
+func (n HashedNode) Commit() (*Point, error) {
+	return n.Commitment(), nil
 }
 
 func (HashedNode) Commitment() *Point {
--- a/tree.go
+++ b/tree.go
@@ -77,7 +77,7 @@ type VerkleNode interface {
 
 	// Commit computes the commitment of the node. The
 	// result (the curve point) is cached.
-	Commit() *Point
+	Commit() (*Point, error)
 
 	// Commitment is a getter for the cached commitment
 	// to this node.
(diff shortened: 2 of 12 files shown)


Apostlex0:
Hey, @gballet could you please review the pr and see if the changes are of any help.

hashednode.go
```diff
@@ -44,14 +44,8 @@ func (HashedNode) Get([]byte, NodeResolverFn) ([]byte, error) {
...
-func (HashedNode) Commit() *Point {
-	// TODO: we should reconsider what to do with the VerkleNode interface and how
-	//       HashedNode fits into the picture, since Commit(), Commitment() and Hash()
-	//	     now panics. Despite these calls must not happen at runtime, it is still
-	//	     quite risky. The reason we end up in this place is because PBSS came quite
-	//	     recently compared with the VerkleNode interface design. We should probably
-	//	     reconsider splitting the interface or find some safer workaround.
-	panic("can not commit a hash node")
+func (n HashedNode) Commit() (*Point, error) {
+	return n.Commitment(), nil
```
gballet: this is a no-op change, and it removes some comments that are useful in case we want to resurrect this repo later.

tree.go
```diff
@@ -823,11 +833,11 @@ func (n *InternalNode) Commit() *Point {
 		minBatchSize := 4
 		if len(nodes) <= minBatchSize {
 			if err := commitNodesAtLevel(nodes); err != nil {
-				// TODO: make Commit() return an error
-				panic(err)
+				return nil, fmt.Errorf("committing nodes at level %d: %w", level, err)
 			}
 		} else {
 			var wg sync.WaitGroup
+			errChan := make(chan error, 1)
```
gballet: this is incorrect, as more than one goroutine can write an error to the channel. If so, multiple goroutines can leak.

gballet (changes requested):
Thanks for the PR. This repo is currently no longer used in geth and will probably be archived soon (although I will keep it alive if it's useful to some people).

In order to merge this, the requested changes have to be applied, but on top of that, extra care needs to be taken that if something goes wrong, there is enough information given to debug this. Which means that info about the curve point needs to be added to the only error that can arise in `commitNodesAtLevel`.

PR #469, opened 2025-11-06
Fix global mutation issue

tudor-malene:
I have been chasing a strange bug that started with nodes getting out of sync when executing a contract (while trying out the pathdb of geth).

It turned out that it only happened if the contract had an array of just the right size to start filling the second half of a verkle node's children. 
After more investigation, it turned out the reason is this global Identity mutation.

I added a test to replicate the issue.


gballet (approved):
Thanks. You caught me in the preparations to devconnect, sorry to have left it drag on. It's a good catch, thank you very much for that.

